PR #1, opened 2019-09-06
Plugin Refactor to use arbitrary state machines

kumare3:
Goal:
 Propeller should support more types of transitions so that plugin
writers can handle cases where it is not possible to get idempotency
from downstream services. Also provide a simplified API that makes
writing plugins easy.

Done:
 - New Node handler interface
 - Implemented Task handler
 - Implemented Dynamic handler
 - modified node executor
 - Unit tests for end to end


pkg/controller/nodes/task/catalog/datacatalog/transformer.go
```diff
@@ -131,16 +134,16 @@ func GenerateArtifactTagName(ctx context.Context, inputs *core.LiteralMap) (stri
 // Get the DataSetID for a task.
 // NOTE: the version of the task is a combination of both the discoverable_version and the task signature.
 // This is because the interfact may of changed even if the discoverable_version hadn't.
-func GenerateDatasetIDForTask(ctx context.Context, task *core.TaskTemplate) (*datacatalog.DatasetID, error) {
-	datasetVersion, err := generateDataSetVersionFromTask(ctx, task)
+func GenerateDatasetIDForTask(ctx context.Context, k catalog.Key) (*datacatalog.DatasetID, error) {
+	datasetVersion, err := generateDataSetVersionFromTask(ctx, k.Type, k.TypedInterface, k.CacheVersion)
```
chanadian: @EngHabu @kumare3 when running this on staging, the `catalog.Key.CacheVersion` is empty.

https://kibana-staging.lyft.net/goto/aa22b6c62a1ff11549c691535110d07c

honnix:
It seems this big change has not been reflected in `sandbox` deployment because it keeps crashing due to configuration structure change. @kumare3

PR #2, opened 2019-09-09
support GCP as auth provider

honnix:
It is already possible to use `kubectl flyte --kubeconfig ~/.kube/config get ...` connecting to a GKE cluster. This PR makes GCP auth supported natively so GKE users will not need to point to kube config manually.


honnix:
@kumare3 This PR is the continuation a previous issue. PTAL and see whether this is too vendor specific that we won't want. Thanks.

kumare3:
Thank you

honnix:
@kumare3 Could you please take another look?

PR #4, opened 2019-09-09
Bump IDL/Plugins versions

wild-endeavor:
Bumping versions to pick up Flyte IDL change deprecating Hive query collections.

https://github.com/lyft/flyteplugins/releases/tag/v0.1.2
https://github.com/lyft/flyteidl/releases/tag/v0.14.0


matthewphsmith:
If this is the only changes in propeller, do we need to bump it at all?

PR #5, opened 2019-09-10
Datacatalog cache

chanadian:
This PR adds the catalog client to cache Task Executions to DataCatalog. 

The steps to cache a task execution are:

1. Create a dataset for the task
2. Create an artifact that represents the execution, along with the artifact data that represents the execution output
3. Tag the artifact with a unique hash of the input values

When retrieving a cached entry:

1. Compute the tag by computing the hash of the input
2. Check if a tagged artifact exists with that hash

Here's how fields in Catalog look like when they are cached:

Every task instance is represented as a DataSet:
```
Dataset {
  project: Flyte project the task was registered in
  domain: Flyte domain for the task execution
  name: flyte_task-<taskName>
  version: <discoverable_version>-<hash(input params)>-<hash(output params)>
// note that a change in the signature will change the hashes and create a new version of the Dataset
}
```

Every task execution is represented as an Artifact in the Dataset above:
```
Artifact {
  id: uuid
  Metadata: [executionName, executionVersion]
}
```

with outputs as ArtifactData:
```
ArtifactData {
  Name: <output-name>
  value: <literal value of the output>
}
```

To retrieve the Artifact, we use the tag associated with the Artifact which is composed of:
```
ArtifactTag {
  Name: flyte_cached-<unique hash of the input values>
}
```


chanadian:
@kumare3 could you please take a look? @surindersinghp reviewed it previously and was OK with it.

The PR adds the catalog client to cache task executions onto Data Catalog. Added tests, documentation and configs. In the future when we move the IDL to FlyteIDL we can remove the old legacy client.

pkg/controller/catalog/datacatalog/datacatalog.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// Tag the artifact since it is the cached artifact
+	tagName, err := transformer.GenerateArtifactTagName(ctx, inputs)
+	if err != nil {
+		logger.Errorf(ctx, "Failed to create tag for artifact %+v, err: %+v", cachedArtifact.Id, err)
+		return err
+	}
+	logger.Debugf(ctx, "Created tag: %v, for task: %v", tagName, task.Id)
+
+	// TODO: We should create the artifact + tag in a transaction when the service supports that
```
kumare3: Why?

chanadian: Because it will avoid race conditions, for example if there are two run of the same task that happen in parallel we may not be tagging the artifact we just created in this service call. It's not a huge deal because either of those artifacts will work, just can be unexpected behavior.

PR #7, opened 2019-09-11
bump-versions of flyteidl and flyteplugins

vsbus:


vsbus:
PTAL @wild-endeavor @migueltol22 @wzpher

PR #8, opened 2019-09-11
github.com/dgrijalva/jwt-go was missing

wild-endeavor:
Could not compile.  Apparently dep ensure doesn't update transitive dependencies.   This just runs `dep ensure -update`.


kumare3:
👍

PR #9, opened 2019-09-13
Adjust Catalog logs

chanadian:
The Catalog client is little too verbose when it errors. There are expected `NotFound` errors that we can downgrade to `debug`. Also the caller of the client logs errors so we don't need to explicitly do it in these cases.


chanadian:
👀 @wild-endeavor

PR #13, opened 2019-09-18
Metrics captured for offending workflows

kumare3:
There are times when one workflow may take a longer than usual in a specific round. We emit a labeled metric so that we can segment metrics by the workflow id. We also try to capture the time spent in plugin invocation.

### DIFF
--- a/pkg/controller/nodes/executor.go
+++ b/pkg/controller/nodes/executor.go
@@ -36,7 +36,8 @@ type nodeMetrics struct {
 	TransitionLatency labeled.StopWatch
 	// Measures the latency between the time a node's been queued to the time the handler reported the executable moved
 	// to running state
-	QueuingLatency labeled.StopWatch
+	QueuingLatency    labeled.StopWatch
+	NodeExecutionTime labeled.StopWatch
 }
 
 type nodeExecutor struct {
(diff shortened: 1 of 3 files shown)


pkg/controller/nodes/executor.go
```diff
@@ -455,6 +456,8 @@ func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, w v1alpha1.Exec
 	switch nodeStatus.GetPhase() {
 	case v1alpha1.NodePhaseNotYetStarted, v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning, v1alpha1.NodePhaseFailing, v1alpha1.NodePhaseRetryableFailure, v1alpha1.NodePhaseSucceeding:
 		logger.Debugf(currentNodeCtx, "Handling node Status [%v]", nodeStatus.GetPhase().String())
+		t := c.metrics.NodeExecutionTime.Start(ctx)
```
EngHabu: Is it possible to measure just the executeNode overhead excluding recursion?

kumare3: yup done, i think the most important thing to capture is data-gather latency. I have added a metric for that

PR #20, opened 2019-10-02
changing the workflow handler to comply with new interface

surindersinghp:
Pending: tests


surindersinghp:
@anandswaminathan not a no-op

PR #24, opened 2019-10-22
Clean up Cache stats

chanadian:
Previously we were returning an error when the cache Tag already exists. This is a valid case when tasks with the same input run in quick succession, so it should not propagate an err.


chanadian:
👀 @EngHabu

PR #25, opened 2019-10-22
Plugin refactor rsrcmgr

kumare3:


pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (rrmm RedisResourceManagerMetrics) GetScope() promutils.Scope {
+	return rrmm.Scope
+}
+
+func (r *RedisResourceManager) getResource(namespace resourcemanager_interface.ResourceNamespace) *Resource {
+	return r.namespacedResourcesMap[namespace]
+}
+
+func (r *RedisResourceManager) pollRedis(ctx context.Context, namespace resourcemanager_interface.ResourceNamespace) {
```
kumare3: migrate to using prometheus metrics gatherer

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Add this registration to the list
+	r.namespacedResourcesMap[namespace] = newResource
+	return nil
+}
+
+func (r *RedisResourceManager) AllocateResource(ctx context.Context, namespace resourcemanager_interface.ResourceNamespace, allocationToken string) (
+	resourcemanager_interface.AllocationStatus, error) {
+
+	namespacedRedisSetKey := r.getNamespacedRedisSetKey(namespace)
```
kumare3: TODO we should check if we can use Redis transactions to implement this as a conditional update

kumare3:
Couple comments,
I would split the ResourceManager and the ResourceNegotiator to differentiate the roles. 
ResourceNegotiator mutates the resourcequotas, but resourcemanager does not allow mutations.
Thus to create a resource manager, we could have a specific method that looks like
NewResourceManager(ResourceNegotiator).

ResourceNegotiator can have an api that returns all currently configured quotas

`GetQuotas() map[...]....`

pkg/controller/nodes/task/handler.go
```diff
@@ -150,17 +153,38 @@ func (t *Handler) setDefault(ctx context.Context, p pluginCore.Plugin) error {
...
-	tSCtx := t.newSetupContext(ctx, sCtx)
+	tSCtx, err := t.newSetupContext(ctx, sCtx)
+	if err != nil {
+		return err
+	}
 
+	// Create a new base resource negotiator
+	resourceManagerConfig := rmConfig.GetResourceManagerConfig()
+	// TODO: check with Ketan / Haytham if it is OK to use the SetupContext's scope here
+	newResourceManagerBuilder, err := resourcemanager.GetResourceManagerBuilderByType(ctx, resourceManagerConfig.ResourceManagerType, tSCtx.MetricsScope())
```
kumare3: i would use the handler's scope

bnsblue:
🔨

PR #27, opened 2019-11-04
Generic type support in Workflow compiler

kumare3:
- For Scalar type Generic, the compiler was missing support. Added this support.

### DIFF
--- a/pkg/compiler/validators/typing_test.go
+++ b/pkg/compiler/validators/typing_test.go
@@ -221,6 +221,22 @@ func TestMapCasting(t *testing.T) {
 		assert.False(t, castable, "{k: Integer} should not be castable to {k: Float}")
 	})
 
+	t.Run("ScalarStructToStruct", func(t *testing.T) {
+		castable := AreTypesCastable(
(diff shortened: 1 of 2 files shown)


matthewphsmith:
You might want to double check the logic for how bindings are constructed with Generic values too. The reason I say this is because:

1. How we implemented a certain legacy type was with a Generic. And it works fine because it is never provided as a workflow input or as a static value. It always go from one task to another and works fine.

2. In our prototype I noticed the failure both when it was a Generic input to a workflow and when I statically set it on a task.

kumare3:
> You might want to double check the logic for how bindings are constructed with Generic values too. The reason I say this is because:
> 
> 1. How we implemented a certain legacy type was with a Generic. And it works fine because it is never provided as a workflow input or as a static value. It always go from one task to another and works fine.
> 2. In our prototype I noticed the failure both when it was a Generic input to a workflow and when I statically set it on a task.

Do you mean to check the runtime behavior? I did, Scalars are pretty simply copied over, by value

PR #30, opened 2019-11-17
Implement a recursive finalize in node executor

EngHabu:
Implement Finalize in Dynamic Node Handler.
The change is more nuanced than this. It changes how Dynamic Nodes fail.

- When a sub-node in a dynamic node fails, the dynamic node will continue to be marked as `Running` in this cycle but change its internal phase to `failing`
- Next cycle, the dynamic node will abort all other sub-nodes that are still running then mark itself as `retryable failure`


pkg/controller/nodes/dynamic/handler.go
```diff
@@ -117,31 +118,39 @@ func (d dynamicNodeTaskNodeHandler) handleDynamicSubNodes(ctx context.Context, n
...
 	newState := ds
+	logger.Infof(ctx, "Dynamic handler.Handle's called with phase %v.", ds.Phase)
 	if ds.Phase == v1alpha1.DynamicNodePhaseExecuting {
-		trns, err = d.handleDynamicSubNodes(ctx, nCtx)
+		trns, newState, err = d.handleDynamicSubNodes(ctx, nCtx, ds)
 		if err != nil {
 			logger.Errorf(ctx, "handling dynamic subnodes failed with error: %s", err.Error())
 			return trns, err
 		}
+	} else if ds.Phase == v1alpha1.DynamicNodePhaseFailing {
```
kumare3: I think this is an excellent addition, but we should probably abort in this case. So Dynamic now has 3 potential steps
New -> Executing (potentially -> Failing in case of failures)
then the abort for the node could propagate abort, if the state is `failing/executing`

kumare3: and ofcourse finalize will be simplified

pkg/controller/nodes/executor.go
```diff
@@ -461,6 +472,59 @@ func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, w v1alpha1.Exec
 	return executors.NodeStatusUndefined, errors.Errorf(errors.IllegalStateError, currentNode.GetID(), "Should never reach here")
 }
 
+func (c *nodeExecutor) FinalizeHandler(ctx context.Context, w v1alpha1.ExecutableWorkflow, currentNode v1alpha1.ExecutableNode) error {
+	nodeStatus := w.GetNodeExecutionStatus(currentNode.GetID())
+	switch nodeStatus.GetPhase() {
+	case v1alpha1.NodePhaseRunning, v1alpha1.NodePhaseFailing, v1alpha1.NodePhaseSucceeding, v1alpha1.NodePhaseRetryableFailure, v1alpha1.NodePhaseQueued:
```
kumare3: I think we should be finalizing only if it is succeeding, failing or retryable failure.
Abort automatically finalizes?

PR #39, opened 2019-12-02
Adding workflow name to workflow crd labels

surindersinghp:
We need this information so we can get pod level usage metrics for workflows. I will be adding a similar change later to plugins repo for array job.

### DIFF
--- a/pkg/compiler/transformers/k8s/workflow.go
+++ b/pkg/compiler/transformers/k8s/workflow.go
@@ -14,7 +14,7 @@ import (
 )
 
 const ExecutionIDLabel = "execution-id"
-const WorkflowIDLabel = "workflow-id"
+const WorkflowNameLabel = "workflow-name"
 
 func requiresInputs(w *core.WorkflowTemplate) bool {
 	if w == nil || w.GetInterface() == nil || w.GetInterface().GetInputs() == nil ||
@@ -181,7 +181,7 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 	if err != nil {
 		errs.Collect(errors.NewWorkflowBuildError(err))
 	}
-	obj.ObjectMeta.Labels[WorkflowIDLabel] = utils.SanitizeLabelValue(WorkflowNameFromID(primarySpec.ID))
+	obj.ObjectMeta.Labels[WorkflowNameLabel] = utils.SanitizeLabelValue(WorkflowNameFromID(primarySpec.ID))
 
 	if obj.Nodes == nil || obj.Connections.DownstreamEdges == nil {
 		// If we come here, we'd better have an error generated earlier. Otherwise, add one to make sure build fails.
--- a/pkg/compiler/transformers/k8s/workflow.go
+++ b/pkg/compiler/transformers/k8s/workflow.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lyft/flytepropeller/pkg/compiler/common"
 	"github.com/lyft/flytepropeller/pkg/compiler/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/lyft/flytepropeller/pkg/utils"
 )
 
 const ExecutionIDLabel = "execution-id"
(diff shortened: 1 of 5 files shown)


pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -171,6 +171,7 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 	if err != nil {
 		errs.Collect(errors.NewWorkflowBuildError(err))
 	}
+	obj.ObjectMeta.Labels[WorkflowIDLabel] = primarySpec.ID
```
EngHabu: There is a limit on the length of label fields, beyond which the creation of the CRD will fail... we should probably add code to guard against that.

EngHabu: Is it really the full `ID`? or just the `Name` part of the ID?

surindersinghp: fixed, it was more to check if this is the right place to inject this.

surindersinghp: thanks for asking. It was project:domain:name, changed it to name. Added a simple test.

surindersinghp:
> Can you summarize the motivation behind the change?
We need this information so we can pod level usage metrics for workflows. I will be adding a similar change later to plugins repo for array job.
@EngHabu

pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -171,6 +181,7 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 	if err != nil {
 		errs.Collect(errors.NewWorkflowBuildError(err))
 	}
+	obj.ObjectMeta.Labels[WorkflowIDLabel] = utils.SanitizeLabelValue(WorkflowNameFromID(primarySpec.ID))
```
EngHabu:
```suggestion
	obj.ObjectMeta.Labels[WorkflowNameLabel] = utils.SanitizeLabelValue(WorkflowNameFromID(primarySpec.ID))
```

pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/lyft/flytepropeller/pkg/compiler/common"
```
EngHabu: Also note, you will need to `dep ensure -update` in flyte admin to pull in this change

pkg/controller/nodes/node_exec_context.go
```diff
@@ -96,9 +100,19 @@ func (e execContext) MaxDatasetSizeBytes() int64 {
...
+	return e.nodeLabels
+}
+
 func newNodeExecContext(_ context.Context, store *storage.DataStore, w v1alpha1.ExecutableWorkflow, node v1alpha1.ExecutableNode, nodeStatus v1alpha1.ExecutableNodeStatus, inputs io.InputReader, maxDatasetSize int64, er events.TaskEventRecorder, tr handler.TaskReader, nsm *nodeStateManager, enqueueOwner func() error) *execContext {
+	md := execMetadata{WorkflowMeta: w}
+	nodeLabels := md.GetLabels()
+	if nodeLabels == nil {
+		nodeLabels = make(map[string]string)
+	}
+	nodeLabels[NodeIDLabel] = utils.SanitizeLabelValue(node.GetID())
```
EngHabu: Should we also only put the node name here? or do you need the full id? the id will include the execution id as well?

pkg/compiler/transformers/k8s/workflow.go
surindersinghp: fixed

pkg/compiler/transformers/k8s/workflow.go
surindersinghp: yep!

pkg/controller/nodes/node_exec_context.go
surindersinghp: it is just the node name.

surindersinghp:
@EngHabu regarding adding taskname to labels in taskHandler instead of plugin executor (in plugins repo).

Labels are encoded in NodeExecutionMetadata which TaskHandler just passes it down to plugins (via tastcontext https://github.com/lyft/flytepropeller/blob/master/pkg/controller/nodes/task/handler.go#L347). NodeExecContext doesnt have set/update label apis. I think its best we add task label in plugins executor and keep task handler simpler.

PR #41, opened 2019-12-06
Remove flytekit version check

EngHabu:


RubenBarragan:
I've approved since after reviewing, it makes sense to the issue we are having. I'd suggest to hold on to have more eyes reviewing that are more familiar with the files.

RubenBarragan:
Thank you for the help @EngHabu. We are testing this change and will update after.

RubenBarragan:
status: we redeployed and tested on our side (Level5) and it's working now. Thanks again.

PR #47, opened 2019-12-20
Adding metrics to capture pod delay in pending state

surindersinghp:


pkg/controller/nodes/task/handler.go
```diff
@@ -257,6 +258,14 @@ func (t Handler) invokePlugin(ctx context.Context, p pluginCore.Plugin, tCtx *ta
 		return nil, regErrors.Wrapf(err, "Invalid transition for plugin [%s]", p.GetID())
 	}
 
+	// Emit the queue latency if the task has just transitioned from Queued to Running.
+	if ts.PluginPhase == pluginCore.PhaseQueued &&
+		(pluginTrns.pInfo.Phase() == pluginCore.PhaseInitializing || pluginTrns.pInfo.Phase() == pluginCore.PhaseRunning) {
```
akhurana001: Isn't pluginTrns only being updated at line 279 ?

surindersinghp: yes, typo. should have been trns.

PR #48, opened 2019-12-20
Adding output-dir for node output for each retry

surindersinghp:
Reworked retry attempt. Now we have 2 different paths- 
- one for input which is called DataDir
- and the other one for output called OutputDir

New output dir is created for every attempt while input dir remains the same through out the node execution.


pkg/controller/nodes/executor.go
```diff
@@ -252,9 +252,9 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		return executors.NodeStatusUndefined, err
 	}
 
-	if len(nodeStatus.GetDataDir()) == 0 {
+	{
 		// Predicate ready, lets Resolve the data
-		dataDir, err := w.GetExecutionStatus().ConstructNodeDataDir(ctx, c.store, node.GetID())
+		dataDir, err := w.GetExecutionStatus().ConstructNodeDataDir(ctx, c.store, node.GetID(), nodeStatus.GetAttempts())
```
akhurana001: can we add unit-tests to validate this ?

pkg/controller/nodes/executor.go
```diff
@@ -252,9 +252,9 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		return executors.NodeStatusUndefined, err
 	}
 
-	if len(nodeStatus.GetDataDir()) == 0 {
+	{
 		// Predicate ready, lets Resolve the data
```
akhurana001: update comment ?

surindersinghp:
@akhurana001 @EngHabu It seems to be working fine for dynamic nodes now. Can you please review it?

PR #51, opened 2019-12-31
Node executor abort to call finalize even on error

wild-endeavor:
In cases where the abort call fails, we should still call finalize as this is the intended behavior of the finalize construct.

### DIFF
--- a/pkg/controller/nodes/executor_test.go
+++ b/pkg/controller/nodes/executor_test.go
@@ -1229,7 +1229,7 @@ func Test_nodeExecutor_abort(t *testing.T) {
 
 	t.Run("abort error calls finalize", func(t *testing.T) {
 		h := &nodeHandlerMocks.Node{}
-		h.On("Abort", mock.Anything, mock.Anything, mock.Anything).Return(errors.New("test error"))
+		h.OnAbortMatch(mock.Anything, mock.Anything, mock.Anything).Return(errors.New("test error"))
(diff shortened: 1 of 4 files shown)


pkg/controller/nodes/executor_test.go
```diff
@@ -1221,3 +1221,51 @@ func Test_nodeExecutor_timeout(t *testing.T) {
...
 }
+
+func Test_nodeExecutor_abort(t *testing.T) {
+	ctx := context.Background()
+	exec := nodeExecutor{}
+	nCtx := &execContext{}
+
+	t.Run("abort error calls finalize", func(t *testing.T) {
+		h := &nodeHandlerMocks.Node{}
+		h.On("Abort", mock.Anything, mock.Anything, mock.Anything).Return(errors.New("test error"))
```
EngHabu: nit:
```suggestion
		h.OnAbortMatch(mock.Anything, mock.Anything, mock.Anything).Return(errors.New("test error"))
```

pkg/controller/nodes/executor_test.go
```diff
@@ -1221,3 +1221,51 @@ func Test_nodeExecutor_timeout(t *testing.T) {
...
+		h.On("FinalizeRequired").Return(true)
```
EngHabu:
```suggestion
		h.OnFinalizeRequired().Return(true)
```

PR #53, opened 2020-01-02
Fix and re-enable resource revision cache

EngHabu:
- [X] Introduce a config-based factory method to control the workflow store type.
- [X] Change WorkflowStore interface to return the committed version of the workflow object from etcd.
- [X] Change resource version caching to only occur when the etcd. update operation results in a resource version change.
- [X] Fix unit tests to simulate updating resource versions.


pkg/controller/workflowstore/resource_version_caching.go
```diff
@@ -55,41 +58,63 @@ func (r *resourceVersionCaching) Get(ctx context.Context, namespace, name string
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	r.updateRevisionCache(ctx, workflow.Namespace, workflow.Name, workflow.ResourceVersion, workflow.Status.IsTerminated())
-	return nil
+	if newWF != nil {
+		// If the update succeeded AND a resource version has changed (indicating the new WF was actually changed),
+		// cache the old
```
wild-endeavor: can you make this comment even more explicit

wild-endeavor:
```suggestion
		// cache the old.  The behavior this code is trying to accomplish is this.  Normally, if the CRD has not changed, 
		// the code will look at the workflow at the normal frequency.  As soon as something has changed, and we get
		// confirmation that we have written the newer workflow to the api server, and receive a different ResourceVersion,
		// we cache the old ResourceVersion number.  This means that we will never process that exact version again 
		// (as long as the cache is up) thus saving us from things like sending duplicate events.
```

PR #55, opened 2020-01-09
Active timeout should account for retry count

surindersinghp:


pkg/compiler/transformers/k8s/node.go
```diff
@@ -50,7 +52,7 @@ func buildNodeSpec(n *core.Node, tasks []*core.CompiledTask, errs errors.Compile
 	// TODO: Active deadline accounts for the retries and queueing delays. using active deadline = execution deadline.
 	var activeDeadline *v1.Duration
 	if timeout != nil {
-		activeDeadline = &v1.Duration{Duration: 2 * timeout.Duration}
+		activeDeadline = &v1.Duration{Duration: time.Duration(n.Metadata.Retries.Retries) * timeout.Duration}
```
akhurana001: 2 * retries

PR #56, opened 2020-01-14
Fix invalid node lookup on abort/finalize

EngHabu:
Subworkflows used in Dynamic Node Handler override GetNodeExecutionStatus function from the parent workflow, however Abort and Finalize were not using that override and were instead using GetExecutionStatus().GetNodeExecutionStatus(). This would only successfully retrieve nodes that are directly in the workflow as opposed to ones that are present as subnodes of another node.


EngHabu:
PTAL @katrogan @bnsblue

PR #57, opened 2020-01-14
fixing output path in launchplan node

surindersinghp:


pkg/controller/nodes/subworkflow/launchplan_test.go
```diff
@@ -204,6 +204,7 @@ func TestSubWorkflowHandler_CheckLaunchPlanStatus(t *testing.T) {
 	mockNodeStatus := &mocks2.ExecutableNodeStatus{}
 	mockNodeStatus.On("GetAttempts").Return(attempts)
 	mockNodeStatus.On("GetDataDir").Return(dataDir)
+	mockNodeStatus.On("GetOutputDir").Return(dataDir)
```
katrogan: should you assert that this is called?

surindersinghp: sure. The unit tests in general need to be beefed up, I will create a separate jira for that.

PR #68, opened 2020-02-14
BugFix: Fix case in which Parent node in a Dynamic nodes finalize is not invoked

kumare3:
Bug: In a Dynamic Node, Finalize for parent node is never invoked. This happens in all cases of a dynamic node. When finalizers are uses, the finalize method becomes extremely important.

Fix: Finalize for parent node is always invoked when dynamic node is finalized
Corollary: Parent node is finalized when dynamic node is finalized, i.e., after all child nodes have completed. Child nodes get finalized as they complete independently.

### DIFF
--- a/pkg/controller/nodes/dynamic/handler.go
+++ b/pkg/controller/nodes/dynamic/handler.go
@@ -191,11 +191,16 @@ func (d dynamicNodeTaskNodeHandler) Abort(ctx context.Context, nCtx handler.Node
 
 // This is a weird method. We should always finalize before we set the dynamic parent node phase as complete?
 func (d dynamicNodeTaskNodeHandler) Finalize(ctx context.Context, nCtx handler.NodeExecutionContext) error {
+	// We should always finalize the parent node success of failure.
+	// If we use the state to decide the finalize then we will never invoke the finalizer for the parent.
+	logger.Infof(ctx, "Finalizing Parent node")
+	if err := d.TaskNodeHandler.Finalize(ctx, nCtx); err != nil {
+		logger.Errorf(ctx, "Failed to finalize Dynamic Nodes Parent.")
+		return err
+	}
+
 	ds := nCtx.NodeStateReader().GetDynamicNodeState()
-	switch ds.Phase {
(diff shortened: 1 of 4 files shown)


pkg/controller/nodes/dynamic/handler.go
```diff
@@ -191,11 +191,16 @@ func (d dynamicNodeTaskNodeHandler) Abort(ctx context.Context, nCtx handler.Node
 
 // This is a weird method. We should always finalize before we set the dynamic parent node phase as complete?
 func (d dynamicNodeTaskNodeHandler) Finalize(ctx context.Context, nCtx handler.NodeExecutionContext) error {
+	// We should always finalize the parent node success of failure.
+	// If we use the state to decide the finalize then we will never invoke the finalizer for the parent.
+	logger.Infof(ctx, "Finalizing Parent node")
+	if err := d.TaskNodeHandler.Finalize(ctx, nCtx); err != nil {
+		logger.Errorf(ctx, "Failed to finalize Dynamic Nodes Parent.")
+		return err
```
EngHabu: Should we still optimistically finalize the could nodes?

kumare3: i dont understand, do you mean in failure scenario. Child nodes are actually finalized everytime they complete. Abort is the scenario where a failure to finalize parent may cause child nodes to not be finalized, (persistent failure). Maybe we could change the order

wild-endeavor: What do you mean by optimistically haytham?

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -191,11 +191,16 @@ func (d dynamicNodeTaskNodeHandler) Abort(ctx context.Context, nCtx handler.Node
 
 // This is a weird method. We should always finalize before we set the dynamic parent node phase as complete?
 func (d dynamicNodeTaskNodeHandler) Finalize(ctx context.Context, nCtx handler.NodeExecutionContext) error {
+	// We should always finalize the parent node success of failure.
```
bnsblue: nit: typo: of -> or

PR #71, opened 2020-02-19
BugFix: Timeout should cause the existing execution to abort and these should be finite

kumare3:
Bug 1; Execution Timeout currently has a bug such that the existing execution is not aborted. Thus the execution continues to run in the background potentially consuming resources.

Bug 2; On a task timeout the UI shows the existing task as running and a new execution begins

Bug 3;  Timeouts could cause infinite retries, as it bypassed the retry logic

Solution
Every timeout is treated like a retryable failure, and every retryable failure enforces an abort to be invoked. Abort was already expected to be idempotent, thus invoking it for a terminated execution should not cause side-effects

+improvements
Finalize Child workflow always, even if finalizing parent fails

Unit tests:
Updated unit tests and add new unit tests


surindersinghp:
Good catch. 

Bug 2; On a task timeout the UI shows the existing task as running and a new execution begins
>> The UI should have shown retryable failure (with msg=Timeout). Weren't we seeing this?

Bug 3; Timeouts could cause infinite retries, as it bypassed the retry logic
>> This didn't happen because node active deadline used to elapse after 2 execution retries (https://github.com/lyft/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L53). And eventually the background execution was killed via owner-reference.

Thanks!

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -191,8 +191,9 @@ func (d dynamicNodeTaskNodeHandler) Abort(ctx context.Context, nCtx handler.Node
 
 // This is a weird method. We should always finalize before we set the dynamic parent node phase as complete?
 func (d dynamicNodeTaskNodeHandler) Finalize(ctx context.Context, nCtx handler.NodeExecutionContext) error {
-	// We should always finalize the parent node success of failure.
-	// If we use the state to decide the finalize then we will never invoke the finalizer for the parent.
+	// We should always finalize the parent node success or failure.
+	// If we use the phase to decide when to finalize in the case where Dynamic node is in phase Executiing
+	// (i.e. child nodes are now being executed) and Finalize is invoked, we will never invoke the finalizer for the parent.
```
EngHabu: Since you are touching this and my comment on the other PR hasn't been addressed :) maybe you can address it here? or in a separate PR that's fine too.
If the finalize() for parent task fails, we should still call finalize for sub nodes. collect the errors and return then (you can take a look at abort() in node handler for an example for collecting errors)... I think a failure in parent task finalize shouldn't stop us from proceeding to ensure we reap as much resources as we can before giving up (assuming system failure persists)

kumare3: Aaah that’s what you meant 😊

kumare3: > Good catch.
> 
> Bug 2; On a task timeout the UI shows the existing task as running and a new execution begins
> 
> > > The UI should have shown retryable failure (with msg=Timeout). Weren't we seeing this?
>>>>> Nope, I am talking about the task executions. They are shown as running. 

> Bug 3; Timeouts could cause infinite retries, as it bypassed the retry logic
> 
> > > This didn't happen because node active deadline used to elapse after 2 execution retries (https://github.com/lyft/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L53). And eventually the background execution was killed via owner-reference.
> >>>> That will not happen, as recently finalizers were enabled for pods. Thus clearing owner reference will not clear the pods as the finalizers are not cleared unless the pods are finalized

> Thanks!

surindersinghp: Got it, thanks!

PR #72, opened 2020-02-20
Implementing project-level and namespace-level quota cap

bnsblue:


pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -184,16 +199,45 @@ func (r *RedisResourceManager) GetID() string {
 	return RedisSetKeyPrefix
 }
 
+func (r *RedisResourceManager) getNamespaceAllocatedCount(ctx context.Context, client RedisClient, namespace pluginCore.ResourceNamespace, allocationToken string) (int, error) {
+	// TODO: redis's SMembers potentially leads to performance degradation at Redis side when there are too many members in the set.
+	// 		 If that happens, we should replace SMembers with Sscan, which needs extra work to deal with duplicated members
+	tokenPrefix, err := extractTokenPrefix(allocationToken)
```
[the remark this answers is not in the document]
bnsblue: I don't think what you described can achieve the functionality we want to have. What we want is counting the number of strings with a specific prefix -- we cannot "build the final allocation token" in this case.

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -184,16 +175,52 @@ func (r *RedisResourceManager) GetID() string {
 	return RedisSetKeyPrefix
 }
 
-func (r *RedisResourceManager) AllocateResource(ctx context.Context, namespace pluginCore.ResourceNamespace, allocationToken string) (
-	pluginCore.AllocationStatus, error) {
+func (r *RedisResourceManager) checkAgainstOneConstraint(_ context.Context, allAllocated []string,
+	constraint ComposedResourceConstraint) bool {
+	var count int64 = 0
+	for _, allocated := range allAllocated {
+		if does := strings.HasPrefix(allocated, constraint.TargetedPrefixString); does {
```
EngHabu:
```suggestion
		if strings.HasPrefix(allocated, constraint.TargetedPrefixString) {
```

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -184,16 +175,52 @@ func (r *RedisResourceManager) GetID() string {
...
+			count++
+		}
+		if count >= constraint.Value {
```
EngHabu: Nit: move this line/check as a function on the constraint type.

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -184,16 +175,52 @@ func (r *RedisResourceManager) GetID() string {
...
+			return false
+		}
+	}
+	return true
+}
+
+func (r *RedisResourceManager) checkAgainstConstraints(ctx context.Context, client RedisClient, resource pluginCore.ResourceNamespace,
+	constraints []ComposedResourceConstraint) (bool, int, error) {
```
EngHabu:
```suggestion
	constraints []ComposedResourceConstraint) (allowed bool, violatedConstraintIndex int, err error) {
```

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
```diff
@@ -203,19 +230,36 @@ func (r *RedisResourceManager) AllocateResource(ctx context.Context, namespace p
...
 
-	size, err := r.client.SCard(string(namespace)).Result()
+	size, err := r.client.SCard(string(namespace))
 	if err != nil {
 		logger.Errorf(ctx, "Error getting size of Redis set %v", err)
 		return pluginCore.AllocationUndefined, err
 	}
 
-	if size > int64(namespacedResource.quota) {
+	if size >= int64(namespacedResource.quota) {
```
EngHabu: nit: should this be represented as a "constraint" as well and use the globalConstraint.IsAllowed(size) check for consistency?

pkg/controller/nodes/task/resourcemanager/resourceconstraints.go
```diff
@@ -0,0 +1,25 @@
+package resourcemanager
+
+import (
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+	pluginCore "github.com/lyft/flyteplugins/go/tasks/pluginmachinery/core"
+)
+
+type ComposedResourceConstraint struct {
```
EngHabu: nit: docs
also, what does the 'composed' mean in the name?
aslo, can you rename the file to match the type name?

pkg/controller/nodes/task/resourcemanager/resourcemanager.go
```diff
@@ -12,23 +12,36 @@ import (
...
+// appending the actual token string to the prefix using '-' separator
+func (t TokenPrefix) append(tokenWithoutPrefix string) Token {
+	return Token(fmt.Sprintf("%s%s%s", t, tokenNamespaceSeparator, tokenWithoutPrefix))
 }
 
-func composeExecutionUrn(id *core.TaskExecutionIdentifier) string {
-	return execUrnPrefix + execUrnSeparator + id.GetNodeExecutionId().GetExecutionId().GetProject() +
-		execUrnSeparator + id.GetNodeExecutionId().GetExecutionId().GetDomain() + execUrnSeparator + id.GetNodeExecutionId().GetExecutionId().GetName()
+// extending the prefix using ':' separator
+func (t TokenPrefix) extend(prefixPart string) TokenPrefix {
```
EngHabu: When do you use extend vs append? the docs are not clear to me of the implications...

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
bnsblue: refactored

pkg/controller/nodes/task/resourcemanager/resourcemanager.go
bnsblue: Refactored. Now I have TokenPrefix.extend() and Token.prepend(). Is it clearer?

pkg/controller/nodes/task/resourcemanager/redis_resourcemanager.go
bnsblue: Refactored

PR #73, opened 2020-02-20
DataCatalog Artifacts over MaxCacheAge will skip cache retrieval

chanadian:
Introduce a Cache Expiration policy that is enforced on the Propeller side for DataCatalog. If the artifacts are older than `MaxCacheAge` the cache lookup will not use the artifact. Instead it will re-run the task and re-cache the new artifact.


chanadian:
@EngHabu I just updated the PR because I think the age comparison be strictly greater than, rather than `>=`

This makes it consistent with something like s3 retention policy: https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lifecycle-mgmt.html

PR #76, opened 2020-02-28
Add backslash to outputDir for outputwriter

tnsetting:
This is an alternative solution to: https://github.com/lyft/flytekit/pull/84. I don't know if this make sense for your guys but this allows us to not have to update all our existing workflows(update flytekit version if the other PR is merged and released). Or this will create problem in AWS setup?

Problem: When flyte propeller set the --output-prefix to {gcs_bucket}/propeller/{workflow_name}-development-{execution_id}/{task_name}/data/0, gsutil write the output.pb to /0 file instead of creating a new folder 0 and upload the file output.pb to the 0 folder. When flyte propeller trying to validate the '0' folder, it could not find output.pb file and fill the task.
Fix: Append / to the output dir fixes the issue

### DIFF
--- a/pkg/controller/nodes/task/taskexec_context.go
+++ b/pkg/controller/nodes/task/taskexec_context.go
@@ -123,7 +123,8 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 		return nil, err
 	}
 
-	ow := ioutils.NewBufferedOutputWriter(ctx, ioutils.NewRemoteFileOutputPaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir()))
+	// Append backslash to the outputDir
+	ow := ioutils.NewBufferedOutputWriter(ctx, ioutils.NewRemoteFileOutputPaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir() + "/"))
 	ts := nCtx.NodeStateReader().GetTaskNodeState()
 	var b *bytes.Buffer
 	if ts.PluginState != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -123,7 +123,8 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 		return nil, err
 	}
 
-	ow := ioutils.NewBufferedOutputWriter(ctx, ioutils.NewRemoteFileOutputPaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir()))
+	// Append backslash to the outputDir
+	ow := ioutils.NewBufferedOutputWriter(ctx, ioutils.NewRemoteFileOutputPaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir()+"/"))
```
honnix: I would prefer the fix being applied to where the output dir was generated originally.

tnsetting: Yes. But the would require some changes to `ConstructReference` which will require some changes in the flytestdlib.

honnix: Then let's do it. :) Fixing here feels like a bandaid really.

tnsetting: :disappointed: I don't have enough time to finish this before my vacation. Can you take over this?

honnix: We can move forward with lyft/flytekit#84 first and I can rework on this PR.

honnix:
@kumare3 Not very sure of the implication. What do you think if we make the deal that dir URL should always end with a trailing slash.

For example, we should do `outputDir, err := c.store.ConstructReference(ctx, dataDir, "0/")` instead. _Of course ConstructReference needs to be modified accordingly.

PR #78, opened 2020-03-05
Fix token marshaling error in fairness V0

bnsblue:


matthewphsmith:
Is the file rename fine and all?

bnsblue:
Yeah, I think so. `make compile`, `make lint`, and `make test_unit` all succeeded. Also I renamed the file using Goland's `context menu->Refactor->Rename`

PR #79, opened 2020-03-05
Fix token format and add unit tests

bnsblue:


pkg/controller/nodes/task/resourcemanager/resourcemanager.go
```diff
@@ -38,7 +38,7 @@ func composeNamespaceScopePrefix(id *core.TaskExecutionIdentifier) TokenPrefix {
 }
 
 func composeExecutionScopePrefix(id *core.TaskExecutionIdentifier) TokenPrefix {
-	return composeNamespaceScopePrefix(id).extend(id.GetNodeExecutionId().GetExecutionId().GetProject())
+	return composeNamespaceScopePrefix(id).extend(id.GetNodeExecutionId().GetExecutionId().GetName())
```
kumare3: name is the id of the execution, these are fundamentally different things. what do you intend to do?

kumare3: ohh i had to expand scope to understand. I see now.

bnsblue: I intend to use the execution id here. This function was copied from `composeProjectScopePrefix` and I forgot to change this from `GetProject()` to `GetName()`

PR #80, opened 2020-03-06
Fixing a malformatted log in resource manager

bnsblue:


kumare3:
how did this not fail linting?

PR #86, opened 2020-03-12
Add interruptible to node and wf

surindersinghp:
# TL;DR
Allow users to mark the WF/Node as interruptible so they can opt-in/out of spot usage. Spot machines are much cheaper than on-demand but have increased probability of interruption.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
The interruptible flag in CRD is plumbed through to task plugin where it take the appropriate action- containers task plugin may chose to run it on interruptible machines.

## Tracking Issue
https://github.com/lyft/flyte/issues/207

## Follow-up issue
_NA_


pkg/apis/flyteworkflow/v1alpha1/workflow.go
```diff
@@ -34,6 +34,8 @@ type FlyteWorkflow struct {
 	// Value must be a positive integer.
 	// +optional
 	ActiveDeadlineSeconds *int64 `json:"activeDeadlineSeconds,omitempty"`
+	// Interruptible value set here is default value used for nodes unless explicitly set at the node level.
+	Interruptible bool `json:"interruptible,omitempty"`
```
surindersinghp: This is default behavior for the nodes (that don't define interruptible for themselves) in this WF. What do we name it- NodesInterruptible?
cc @migueltol22 @EngHabu @kumare3

EngHabu: Maybe add a `NodeDefaults` struct and add `Interruptible` in there?

pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -318,6 +318,7 @@ type ExecutableNode interface {
 	GetRetryStrategy() *RetryStrategy
 	GetExecutionDeadline() *time.Duration
 	GetActiveDeadline() *time.Duration
+	IsInterruptible() *bool
```
kumare3: should this be a pointer? Doesnt absence of this flag automatically imply a `false`?

kumare3: you can always omitempty, so you dont need to make it a pointer

pkg/apis/flyteworkflow/v1alpha1/workflow.go
```diff
@@ -47,6 +49,11 @@ type FlyteWorkflow struct {
 	DataReferenceConstructor storage.ReferenceConstructor `json:"-"`
 }
 
+type NodeDefaults struct {
+	// Default behaviour for Interruptible for nodes unless explicitly set at the node level.
+	Interruptible bool `json:"interruptible,omitempty"`
```
kumare3: so this is not a pointer and the previous one is. I prefer this though

pkg/controller/config/config.go
```diff
@@ -121,8 +122,9 @@ type WorkqueueConfig struct {
 
 // configuration for a node
 type NodeConfig struct {
-	DefaultDeadlines                DefaultDeadlines `json:"default-deadlines,omitempty" pflag:",Default value for timeouts"`
-	MaxNodeRetriesForSystemFailures int64            `json:"max-node-retries-system-failures" pflag:"2,Maximum number of retries per node for node failure due to infra issues"`
+	DefaultDeadlines               DefaultDeadlines `json:"default-deadlines,omitempty" pflag:",Default value for timeouts"`
+	MaxNodeRetriesOnSystemFailures int64            `json:"max-node-retries-system-failures" pflag:"2,Maximum number of retries per node for node failure due to infra issues"`
+	InterruptibleFailureThreshold  int64            `json:"interruptible-failure-threshold" pflag:"1, number of failures for a node to be still considered interruptible'"`
```
kumare3: lets reword the description. Number of interruptions after which the node is marked as failed?

pkg/controller/nodes/node_exec_context.go
```diff
@@ -152,6 +160,17 @@ func (c *nodeExecutor) newNodeExecContextDefault(ctx context.Context, w v1alpha1
 		return nil
 	}
 
+	interrutible := w.IsInterruptible()
+	if n.IsInterruptible() != nil {
+		interrutible = *n.IsInterruptible()
+	}
+
+	// a node is not considered interruptible if the system failures have exceeded the configured threshold
+	if interrutible && s.GetSystemFailures() >= c.interruptibleFailureThreshold {
```
kumare3: so interruptions are clubbed with system failures is it?

kumare3: is it because there is no specific error that is observed during interruptions?

pkg/apis/flyteworkflow/v1alpha1/iface.go
surindersinghp: Required Behavior: User defines interruptible for a WF which means all the compute from that WF is interruptible. However, this behavior can be overridden by individual nodes.
WF: 
   bool interruptible
      Default value is False: absence of this parameter defaults to false.
Node: 
   bool *interruptible
      In some cases users may actually want to override the interruptible behavior for the node to false. Bool will not work as we will not be able to differentiate absence from false because absence means “use the value defined for the WF, I don’t wish to override"

pkg/controller/config/config.go
surindersinghp: After the threshold is hit, we don't mark the node as failed but don't consider the node as interruptible (eligible for spot). This is to upper bound the number of system failures due to interruption for a particular node.

pkg/controller/nodes/node_exec_context.go
surindersinghp: Yes :(

Unfortunately, we can distinguish kubelet failures from node interruptions or manual node deletions.

pkg/apis/flyteworkflow/v1alpha1/iface.go
EngHabu: You can also make it an enum... this works fine though..

pkg/controller/config/config.go
EngHabu:
```suggestion
	InterruptibleFailureThreshold  int64            `json:"interruptible-failure-threshold" pflag:"1,number of failures for a node to be still considered interruptible'"`
```

pkg/controller/nodes/node_exec_context.go
EngHabu: Apparently with a bit more work, we can: https://github.com/kube-aws/kube-spot-termination-notice-handler
I've not google much but either this or probably another project or we can just modify it to set a label on the pod instead or some other way to mark as being interrupted...

pkg/controller/nodes/executor.go
```diff
@@ -32,12 +32,13 @@ import (
...
-	ResolutionFailure  labeled.Counter
-	InputsWriteFailure labeled.Counter
-	TimedOutFailure    labeled.Counter
+	Scope                   promutils.Scope
+	FailureDuration         labeled.StopWatch
+	SuccessDuration         labeled.StopWatch
+	ResolutionFailure       labeled.Counter
+	InputsWriteFailure      labeled.Counter
+	TimedOutFailure         labeled.Counter
+	InterruptedThresholdHit labeled.Counter
```
EngHabu: Should we add a metric for just all Interrupted events? even if they don't hit the threshold I mean...

pkg/controller/nodes/node_exec_context.go
surindersinghp: We are actually planning to do this in V1 where we pass term signal to the containers as well because that is some work (cloudwatch listener -> ds -> easy support in SDK through a callback).
For V0, we want it to be a simple system error with some info in description.

pkg/controller/nodes/executor.go
surindersinghp: Unfortunately, we are counting system errors towards interrupted threshold. 

However, we are capturing interrupted metrics from kube-state-metrics (not realtime though, infrasepnd model): pod_label_info X pod_info X node_info.

surindersinghp: I will add a counter for system failures.

pkg/apis/flyteworkflow/v1alpha1/iface.go
surindersinghp: Yes, I am fine either ways. Any hard preference?

pkg/controller/nodes/executor.go
surindersinghp: added a counter

pkg/controller/nodes/executor.go
```diff
@@ -38,6 +38,9 @@ type nodeMetrics struct {
 	ResolutionFailure  labeled.Counter
 	InputsWriteFailure labeled.Counter
 	TimedOutFailure    labeled.Counter
+	SystemFailure      labeled.Counter
```
EngHabu: We already have FailureLatency metric, can we just tag that with System or not?

surindersinghp: The FailureDuration metric which we have today measures end-2-end latency from start of the node to when its actually marked failed. To achieve that, its emitted when we change the phase from Failing to Failed ... basically in the next round and at that time, we dont have error information available in nodestatus. I do feel that we should rework some of that logic.  

I can set up some time next week to talk about it. Please let me know @EngHabu

EngHabu: Ic... Can we then rename this metric to Failure and have a label of failure type and set it to either System or User before Incrementing?

surindersinghp: Setting labels during increment is tricky. 

In the current model (promutils in stdlib), Labeled metrics get label values from context and the keys for those labels are defined-once through SetMetricKeys. That is being set in main.go with values: ProjectKey, DomainKey, WorkflowIDKey, TaskIDKey. So when we incr(), we cannot say system/user/etc unless we add errorKind key to the predefined labels in main.go which I dont want to do :(. @EngHabu  

I dont like this structure but I dont want to work on those issues in this PR.

EngHabu: I understand. Can you get rid of the systemFailures counter then, file an issue to support that in stdlib and assign it to me?

surindersinghp: Done. Thanks @EngHabu

PR #90, opened 2020-03-16
[wip] a regexp used to filter namespaces

honnix:
# TL;DR
When watching "all" namespaces, a regexp filter can be used to narrow down the scope.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added (partially)

## Complete description
**This PR now depends on https://github.com/lyft/flytestdlib/pull/58.**

It is already possible to watch one specific namespace, however it seems not possible to watch
multiple ones according to https://github.com/kubernetes/client-go/issues/580

This PR added a new config (regexp) that will be used to ensure only workflow from interested namespace(s) is processed: informer callback and garbage collector.

This is probably the first time I wrote this much `go` code, please help me improve. Also I haven't spent much time working on `controller` test cases (which seems to be missing heavily).

## Tracking Issue
NA (will create one better context/use case is needed)

## Follow-up issue
NA

### DIFF
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -59,7 +59,7 @@ type Config struct {
 	WorkflowReEval      config.Duration      `json:"workflow-reeval-duration" pflag:"\"30s\",Frequency of re-evaluating workflows"`
 	DownstreamEval      config.Duration      `json:"downstream-eval-duration" pflag:"\"60s\",Frequency of re-evaluating downstream tasks"`
 	LimitNamespace      string               `json:"limit-namespace" pflag:"\"all\",Namespaces to watch for this propeller"`
-	NamespaceFilter     string               `json:"namespace-filter" pflag:",Regular expression to apply fine grain filter when watching all namespaces"`
+	NamespaceFilter     config.Regexp        `json:"namespace-filter" pflag:",Regular expression to apply fine grain filter when watching all namespaces"`
 	ProfilerPort        config.Port          `json:"prof-port" pflag:"\"10254\",Profiler port"`
 	MetadataPrefix      string               `json:"metadata-prefix,omitempty" pflag:",MetadataPrefix should be used if all the metadata for Flyte executions should be stored under a specific prefix in CloudStorage. If not specified, the data will be stored in the base container directly."`
 	Queue               CompositeQueueConfig `json:"queue,omitempty" pflag:",Workflow workqueue configuration, affects the way the work is consumed from the queue."`
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -59,6 +59,7 @@ type Config struct {
 	WorkflowReEval      config.Duration      `json:"workflow-reeval-duration" pflag:"\"30s\",Frequency of re-evaluating workflows"`
 	DownstreamEval      config.Duration      `json:"downstream-eval-duration" pflag:"\"60s\",Frequency of re-evaluating downstream tasks"`
 	LimitNamespace      string               `json:"limit-namespace" pflag:"\"all\",Namespaces to watch for this propeller"`
+	NamespaceFilter     string               `json:"namespace-filter" pflag:",Regular expression to apply fine grain filter when watching all namespaces"`
 	ProfilerPort        config.Port          `json:"prof-port" pflag:"\"10254\",Profiler port"`
 	MetadataPrefix      string               `json:"metadata-prefix,omitempty" pflag:",MetadataPrefix should be used if all the metadata for Flyte executions should be stored under a specific prefix in CloudStorage. If not specified, the data will be stored in the base container directly."`
 	Queue               CompositeQueueConfig `json:"queue,omitempty" pflag:",Workflow workqueue configuration, affects the way the work is consumed from the queue."`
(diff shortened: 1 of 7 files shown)


pkg/controller/config/config.go
```diff
@@ -59,6 +59,7 @@ type Config struct {
 	WorkflowReEval      config.Duration      `json:"workflow-reeval-duration" pflag:"\"30s\",Frequency of re-evaluating workflows"`
 	DownstreamEval      config.Duration      `json:"downstream-eval-duration" pflag:"\"60s\",Frequency of re-evaluating downstream tasks"`
 	LimitNamespace      string               `json:"limit-namespace" pflag:"\"all\",Namespaces to watch for this propeller"`
+	NamespaceFilter     string               `json:"namespace-filter" pflag:",Regular expression to apply fine grain filter when watching all namespaces"`
```
EngHabu: I would ideally create a type like [this](https://github.com/lyft/flytestdlib/blob/master/config/url.go) to be able to get config validation (using `propeller config validate`) before actually running propeller...

EngHabu:
```suggestion
	NamespaceFilterRegex     string               `json:"namespace-filter" pflag:",Regular expression to apply fine grain filter when watching all namespaces"`
```

pkg/controller/controller.go
```diff
@@ -53,9 +54,10 @@ type Controller struct {
 	workflowStore       workflowstore.FlyteWorkflow
 	// recorder is an event recorder for recording Event resources to the
 	// Kubernetes API.
-	recorder      record.EventRecorder
-	metrics       *metrics
-	leaderElector *leaderelection.LeaderElector
+	recorder         record.EventRecorder
+	metrics          *metrics
+	leaderElector    *leaderelection.LeaderElector
+	namespaceFilterR *regexp.Regexp
```
EngHabu:
```suggestion
	namespaceFilterRegex *regexp.Regexp
```

pkg/controller/controller.go
```diff
@@ -220,12 +238,24 @@ func New(ctx context.Context, cfg *config.Config, kubeclientset kubernetes.Inter
 		wfLauncher = launchplan.NewFailFastLaunchPlanExecutor()
 	}
 
+	namespaceFilterR := func() *regexp.Regexp {
+		if cfg.NamespaceFilter != "" {
+			r, err := regexp.Compile(cfg.NamespaceFilter)
+			if err != nil {
+				logger.Warningf(ctx, "[%s] is not a valid regular expression, skipping namespace filtering. Error : [%v]", cfg.NamespaceFilter, err)
```
EngHabu: This should return an error and not try to recover from bad config errors.

pkg/controller/controller.go
```diff
@@ -220,12 +238,24 @@ func New(ctx context.Context, cfg *config.Config, kubeclientset kubernetes.Inter
 		wfLauncher = launchplan.NewFailFastLaunchPlanExecutor()
 	}
 
+	namespaceFilterR := func() *regexp.Regexp {
```
EngHabu: Is there a reason for this predicate?

pkg/controller/controller.go
```diff
@@ -220,12 +238,24 @@ func New(ctx context.Context, cfg *config.Config, kubeclientset kubernetes.Inter
...
+		if cfg.NamespaceFilter != "" {
+			r, err := regexp.Compile(cfg.NamespaceFilter)
```
EngHabu: If you opt for the config type (first comment), you are guaranteed a valid regex at this point...

pkg/controller/controller.go
honnix: I had hard time dealing with `nil`. What I want was:

```
a := nil

if b {
    a = "something not nil"
}
```

It's my bad `go` skill I believe.

honnix:
@EngHabu Yeah, client side filtering is definitely not ideal. It's good to know that you have been experimenting with a replacement. I remember @kumare3 mentioned that when we talked last time. It's been a while and I probably forgot about the details. 😅 

Our immediate use case is actually not for sharding, partitioning or in general for performance sake.

Imagine we have a production setup with all the components running and propeller watching `all`; we would like to start a propeller locally that can utilise everything in the production (admin, console, k8s cluster, etc.) but only watches a debug namespace. With the production propeller watching `all`, we cannot route the workflow execution to the local propeller only.

pkg/controller/controller.go
honnix: Yeah that might be better. I wasn't so sure whether this should be treated as failure. This could be done better as suggested by https://github.com/lyft/flytepropeller/pull/90#discussion_r393113304

pkg/controller/config/config.go
honnix: Will fix in a commit for code alignment sake.

pkg/controller/controller.go
honnix: Will fix in a commit for code alignment sake.

honnix:
That being said, please do not consider this as must to have feature. The PR was more my exercise while reading through propeller code base.

pkg/controller/config/config.go
honnix: Thanks for the suggestion. I tried it in https://github.com/lyft/flytestdlib/pull/58.

honnix:
It would be a lot better this filtering is done on server side than client. Closing this PR for now.

PR #94, opened 2020-03-19
Bug/Feature: Dynamic parent node should be finalized earlier

kumare3:
# TL;DR
This is not really a bug, but it so happens that the Parent node is stuck in terminating until the entire dynamic node is finalized. Which may take hours depending on the subnodes. This fix enable s Dynamic node to be released and allow termination as soon as it is complete

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Dynamic node now has an intermediate state that especially handles finalization of the parent node

## Tracking Issue
https://github.com/lyft/flyte/issues/209

## Follow-up issue
NA


surindersinghp:
looks good.

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -129,6 +129,8 @@ func (d dynamicNodeTaskNodeHandler) handleDynamicSubNodes(ctx context.Context, n
 	return trns, newState, nil
 }
 
+// The State machine for a dynamic node is as follows
```
EngHabu: Did you mean to complete this paragraph ?

kumare3: ohh ya, fragmented the brain

PR #95, opened 2020-03-19
Adding metrics for User vs system errors

anandswaminathan:
This PR propagates metrics for errors from the tasks - user/system/unknown.

Based on the execErr.GetKind(), we emit metrics if the error was user/system to gain more insight into the executions.


pkg/controller/nodes/executor.go
```diff
@@ -403,7 +406,18 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		logger.Errorf(ctx, "failed Execute for node. Error: %s", err.Error())
 		return executors.NodeStatusUndefined, err
 	}
-
+	execErr := p.GetErr()
+	if execErr != nil {
```
surindersinghp: line duplicated

pkg/controller/nodes/task/handler.go
```diff
@@ -606,8 +606,8 @@ func New(ctx context.Context, kubeClient executors.Client, client catalog.Client
 			catalogPutSuccessCount: labeled.NewCounter("discovery_put_success_count", "Discovery Put success count", scope),
 			catalogPutFailureCount: labeled.NewCounter("discovery_put_failure_count", "Discovery Put failure count", scope),
 			catalogGetFailureCount: labeled.NewCounter("discovery_get_failure_count", "Discovery Get faillure count", scope),
-			pluginExecutionLatency: labeled.NewStopWatch("plugin_exec_latecny", "Time taken to invoke plugin for one round", time.Microsecond, scope),
```
surindersinghp: revert

anandswaminathan: why? `plugin_exec_latecny ` spelling

surindersinghp: sorry!

pkg/controller/nodes/executor.go
```diff
@@ -403,7 +406,16 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		logger.Errorf(ctx, "failed Execute for node. Error: %s", err.Error())
 		return executors.NodeStatusUndefined, err
 	}
-
+	execErr := p.GetErr()
```
EngHabu: I think putting this code here will cause double counting. because -for example- we will go here when the node fails, then it moves to failing, then we will go here again after finalize...

EngHabu: Also, do we want to mark spot instance disappearing as a system error? maybe a question to @surindersinghp

anandswaminathan:
PTAL @surindersinghp @EngHabu

pkg/controller/nodes/executor.go
surindersinghp: I think yes for now. And as we are able to pinpoint interruptions, we can add one more type (or via label).

PR #97, opened 2020-03-23
[Bug Fix] LastAttemptStartedAt was not set if state NodePhase is skipped

surindersinghp:
LastAttemptStartedAt is not always set. Make sure `LastAttemptStartedAt ` is set and be extra careful that `LastAttemptStartedAt ` is checked for nil before using it for metrics.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ X] Code completed
 - [X ] Smoke tested
 - [ ] Unit tests added


anandswaminathan:
+1

pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -403,6 +403,9 @@ func (in *NodeStatus) UpdatePhase(p NodePhase, occurredAt metav1.Time, reason st
 		if in.StartedAt == nil {
 			in.StartedAt = &n
 		}
+		if in.LastAttemptStartedAt == nil {
```
EngHabu: This means it'll only be set once, right? shouldn't this be set when we update the status to running again?

surindersinghp: Correcto. I think its being reset after we retry. If not then we should do that but not remove null check.

surindersinghp: https://github.com/lyft/flytepropeller/blob/master/pkg/controller/nodes/executor.go#L392

PR #98, opened 2020-03-24
Emit error metrics for all failures

surindersinghp:
We found metrics not reported. On further digging, we realized, that status is cleared in the Retry case, before we were using. Also in cases, few extra checks regarding node transitions prevented metrics from getting emitted.

## Type
 - [X ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ X] Code completed
 - [ X] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/pkg/controller/nodes/executor.go
+++ b/pkg/controller/nodes/executor.go
@@ -398,7 +398,7 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		return executors.NodeStatusFailed(fmt.Errorf(nodeStatus.GetMessage())), nil
 	}
 
-	// case v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning, v1alpha1.NodePhaseRetryableFailure:
+	// case v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning:
 	logger.Debugf(ctx, "node executing, current phase [%s]", currentPhase)
 	defer logger.Debugf(ctx, "node execution completed")
 	p, err := c.execute(ctx, h, nCtx, nodeStatus)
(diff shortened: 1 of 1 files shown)


pkg/controller/nodes/executor.go
```diff
@@ -406,8 +406,9 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		logger.Errorf(ctx, "failed Execute for node. Error: %s", err.Error())
 		return executors.NodeStatusUndefined, err
 	}
+
 	execErr := p.GetErr()
-	if execErr != nil && p.GetPhase() == handler.EPhaseRetryableFailure && nodeStatus.GetLastAttemptStartedAt() != nil {
+	if execErr != nil && nodeStatus.GetLastAttemptStartedAt() != nil {
```
anandswaminathan: Add a comment here:

// Err object should not be populated for other phase, and we only reach here for v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning

Also comment on lint 401 is incorrect. Remove v1alpha1.NodePhaseRetryableFailure in line 401

pkg/controller/nodes/executor.go
```diff
@@ -398,23 +398,33 @@ func (c *nodeExecutor) handleNode(ctx context.Context, w v1alpha1.ExecutableWork
 		return executors.NodeStatusFailed(fmt.Errorf(nodeStatus.GetMessage())), nil
 	}
 
-	// case v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning, v1alpha1.NodePhaseRetryableFailure:
+	// case v1alpha1.NodePhaseQueued, v1alpha1.NodePhaseRunning:
 	logger.Debugf(ctx, "node executing, current phase [%s]", currentPhase)
 	defer logger.Debugf(ctx, "node execution completed")
 	p, err := c.execute(ctx, h, nCtx, nodeStatus)
```
anandswaminathan: Execute clears the `LastAttemptStartedAt` for `EPhaseRetryableFailure`

PR #100, opened 2020-03-31
Fix exponent overflow

EngHabu:
# TL;DR
Backoff controller has two issues: 1) backoff() can overflow the exponent value leading to 0 backoff duration all the time. 2) There is a race condition in updating fields of the simple backoff handler. this PR should address both.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/233

### DIFF
--- a/pkg/controller/nodes/task/backoff/handler.go
+++ b/pkg/controller/nodes/task/backoff/handler.go
@@ -5,10 +5,11 @@ import (
 	"math"
 	"regexp"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/lyft/flyteplugins/go/tasks/errors"
+	stdAtomic "github.com/lyft/flytestdlib/atomic"
 	stdErrors "github.com/lyft/flytestdlib/errors"
 	"github.com/lyft/flytestdlib/logger"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


bnsblue:
Not sure if a lock is really needed in backoff handler because the race condition should be benign (increment and reset to 0 only). But if we don't see much performance hit we can keep it just to be safe.

pkg/controller/nodes/task/backoff/handler.go
```diff
@@ -24,33 +26,40 @@ var (
 type SimpleBackOffBlocker struct {
 	Clock              clock.Clock
 	BackOffBaseSecond  int
-	BackOffExponent    int
-	NextEligibleTime   time.Time
 	MaxBackOffDuration time.Duration
+
+	// Mutable fields
+	BackOffExponent  stdAtomic.Uint32
+	NextEligibleTime atomic.Value
```
kumare3: Ideally we should implement the same interface as the stdAtomic for time, if we can
`NextEligibleTime atomic.Value` Then we can migrate it to stdlib at somepoint

PR #102, opened 2020-04-03
Refactor to remove Workflow from Node executor

kumare3:
# TL;DR
This PR does not introduce Single Task execution, but is a precursor. Even if we do not implement Single Task Execution, this PR improves the hygiene of the code. One of the largest changes is to change from classical Golang Mockery's way of mocking methods to use @EngHabu's wonderful fork that generates type safe methods (albeit this change makes the PR looks bigger)

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The PR tries to address the following problems
 - Propellers NodeExecutor accepts a workflow as an argument. When executing a node only needs the node information. 
 - The Workflow is passed into every node handler, this is also an oversight. Makes it harder to write comprehensive unit tests and the coupling prevents future extensions
 - The default entrypoint for node executor is called RecursiveNodeHandler (actually should be traversal node handler). This needs the DAGStructure. But there are cases in which a node can just be executed without having to traverse down the graph.
 - ContextualWorkflow is a hard concept to understand!

Solution:
 - We are now adding 3 new structures
 1. DAGStructure
    This is a meta structure that only captures the structural information of the Direct Acyclic Graph (i.e. the relationships between the nodes)
 2. NodeLookup
     this is a structure that allows users of this structure to look up a node or the status of a specific node within the current sub-graph
 3. ExecutionContext
     This structure essentially holds meta information about the execution like the Execution ID etc. It also provides methods to lookup tasks and Subworkflows that could be either defined statically or generated dynamically.

## Tracking Issue
https://github.com/lyft/flyte/issues/221

## Follow-up issue
1. Rename this method to DAGTraversalHandleNode (accepts a DAGStructure along-with) the remaining arguments
2. Create a new method called HandleNode (part of the interface) (remaining all args as the previous method, but no DAGStructure
3. Additionally both methods will receive inputs reader
4. The Downstream nodes handler will Resolve the Inputs
5. the method will delegate all other node handling to HandleNode.
6. Thus we can get rid of SetInputs for StartNode as well

All of this will be done part of the Same issue #221 (cannot be closed)


pkg/controller/nodes/branch/handler.go
```diff
@@ -154,19 +158,47 @@ func (b *branchHandler) Abort(ctx context.Context, nCtx handler.NodeExecutionCon
...
+	}
+
+	finalNodeID := branchNodeState.FinalizedNodeID
+	branchTakenNode, ok := nCtx.ContextualNodeLookup().GetNode(*finalNodeID)
+	if !ok {
+		logger.Errorf(ctx, "Downstream node [%v] not found", *finalNodeID)
+		return nil
+	}
+
+	// TODO @kumare before merging, we should have a non DAG structure passed in here
```
EngHabu, line 199: ?

pkg/controller/nodes/branch/handler.go
```diff
@@ -154,19 +158,47 @@ func (b *branchHandler) Abort(ctx context.Context, nCtx handler.NodeExecutionCon
...
 
 	finalNodeID := branchNodeState.FinalizedNodeID
-	branchTakenNode, ok := w.GetNode(*finalNodeID)
+	branchTakenNode, ok := nCtx.ContextualNodeLookup().GetNode(*finalNodeID)
 	if !ok {
 		logger.Errorf(ctx, "Downstream node [%v] not found", *finalNodeID)
 		return nil
 	}
 
+	// TODO @kumare before merging, we should have a non DAG structure passed in here
```
EngHabu, line 167: ?

pkg/controller/nodes/branch/handler.go
```diff
@@ -88,28 +81,40 @@ func (b *branchHandler) Handle(ctx context.Context, nCtx handler.NodeExecutionCo
...
+	nl := nCtx.ContextualNodeLookup()
+
+	return b.HandleBranchNode(ctx, branchNode, nCtx, nl)
+}
+
 func (b *branchHandler) recurseDownstream(ctx context.Context, nCtx handler.NodeExecutionContext, nodeStatus v1alpha1.ExecutableNodeStatus, branchTakenNode v1alpha1.ExecutableNode) (handler.Transition, error) {
-	w := nCtx.Workflow()
-	downstreamStatus, err := b.nodeExecutor.RecursiveNodeHandler(ctx, w, branchTakenNode)
+	// There is no DAGStructure for the branch nodes
+	// TODO @kumare before merging, we should have a non DAG structure passed in here
```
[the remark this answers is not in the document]
kumare3: yes. good catch :(

pkg/controller/nodes/branch/handler.go
kumare3: Done!

pkg/controller/nodes/branch/handler.go
kumare3: Done!

pkg/controller/nodes/branch/handler.go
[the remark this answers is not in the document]
kumare3: Done

kumare3:
@EngHabu ping again

PR #103, opened 2020-04-03
Retrieve launch plan interfaces from admin

wild-endeavor:
# TL;DR
When Propeller comes across a launch plan node in a `DynamicJobSpec` it will hit Admin to retrieve the interface for the LP to do the interface check.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Please see the issue linked below and also the SDK PR https://github.com/lyft/flytekit/pull/92 for more information.  A sample dynamic job spec object has been uploaded here as well.  Please see the text file for the type of dynamic job spec this PR is meant to support.
[dynamic_job_spec.txt](https://github.com/lyft/flytepropeller/files/4430252/dynamic_job_spec.txt)

* Added a `GetLaunchPlan` function to a `launchplan/Reader` interface which sits alongside the `launchplan/Executor` interface.  Admin client wrapper now satisfies both interfaces.
* Added a call to that function in the dynamic job handler `buildContextualDynamicWorkflow` function.

## Tracking Issue
https://github.com/lyft/flyte/issues/139

## Follow-up issue
https://github.com/lyft/flyte/issues/246
This PR will be deprecated upon completion of this issue.


kumare3:
What is this for? Dynamic LP?

wild-endeavor:
Yes.

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -377,8 +379,19 @@ func (d dynamicNodeTaskNodeHandler) buildContextualDynamicWorkflow(ctx context.C
...
+	requirements, err := compiler.GetRequirements(wf, djSpec.Subworkflows)
+	if err != nil {
+		return nil, true, err
+	}
+
+	launchPlanInterfaces, err := d.getLaunchPlanInterfaces(ctx, requirements.GetRequiredLaunchPlanIds())
+	if err != nil {
+		return nil, true, err
+	}
+
```
EngHabu, line 393: Mind adding a TODO to get the tasks we don't already have from Admin as well...

pkg/compiler/admin.go
```diff
@@ -0,0 +1,37 @@
+package compiler
+
+import (
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+)
+
+// Should we call this something else? It can be used more generally if ever necessary.
```
EngHabu, line 8: Historically: I wanted to make both tasks and launchplan  parameters in the compiler to take InterfaceProviders... but then launch plans diverged  (with expected* stuff) maybe there is still hope?

wild-endeavor: I can't think of a good name... InterfaceProviderProvider was the best I could come up with, so I'm leaving it as is for now.

pkg/controller/nodes/dynamic/handler.go
wild-endeavor: Yup.

pkg/controller/nodes/subworkflow/launchplan/launchplan.go
```diff
@@ -31,6 +31,9 @@ type Executor interface {
 	// Kill a remote execution
 	Kill(ctx context.Context, executionID *core.WorkflowExecutionIdentifier, reason string) error
 
+	// Get the definition of a launch plan. This is primarily used to ensure all the TypedInterfaces match up before actually executing.
+	GetLaunchPlan(ctx context.Context, launchPlanRef *core.Identifier) (*admin.LaunchPlan, error)
```
kumare3, line 35: I think we should create a separate interface for this. And then pass that interface to Dynamic task. This is weird coupling - subworkflow only ever launches a launchplan, having this method is confusing and in the future may lead to weird uses. But having a separate interface will prevent these problems, and also make uses cases clear

pkg/controller/nodes/subworkflow/launchplan/admin.go
```diff
@@ -97,6 +97,26 @@ func (a *adminLaunchPlanExecutor) GetStatus(ctx context.Context, executionID *co
 	return item.ExecutionClosure, item.SyncError
 }
 
+func (a *adminLaunchPlanExecutor) GetLaunchPlan(ctx context.Context, launchPlanRef *core.Identifier) (*admin.LaunchPlan, error) {
```
kumare3, line 100: Having this method here is fine, but having a separate interface for this method is useful to differentiate the use cases and usage paths

pkg/compiler/admin.go
```diff
@@ -0,0 +1,36 @@
+package compiler
+
+import (
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+)
+
+// This object is meant to satisfy github.com/lyft/flytepropeller/pkg/compiler/common.InterfaceProvider
```
kumare3, line 8: This comment does not answer why this is needed here?

kumare3:
Thank you for the changes.

pkg/compiler/admin.go
wild-endeavor: We need a struct to satisfy the interface because the function that we pass it to, `compiler.CompileWorkflow` takes the `common.InterfaceProvider` interface.

PR #104, opened 2020-04-06
Add unlabeled metrics for workflow success and failure durations

EngHabu:
# TL;DR
Emit unlabeled metrics for Workflow success and Failure duration metrics. This enables computing overall system reliability metrics without having to sum individual wf metrics.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added


surindersinghp:
Why do we need to emit a separate metric for aggregation?

EngHabu:
For quantiles, we can't aggregates metrics once the client side aggregation is done. We need this to be able to look at the aggregate quantiles.

surindersinghp:
> For quantiles, we can't aggregates metrics once the client side aggregation is done. We need this to be able to look at the aggregate quantiles.

make sense. thanks!

PR #106, opened 2020-04-07
[Ignore] PR into PR #103

wild-endeavor:


pkg/controller/controller.go
```diff
@@ -199,26 +199,26 @@ func newK8sEventRecorder(ctx context.Context, kubeclientset kubernetes.Interface
 func New(ctx context.Context, cfg *config.Config, kubeclientset kubernetes.Interface, flytepropellerClientset clientset.Interface,
 	flyteworkflowInformerFactory informers.SharedInformerFactory, kubeClient executors.Client, scope promutils.Scope) (*Controller, error) {
 
-	var wfLauncher launchplan.Executor
+	var launchPlanActor launchplan.AdminAccess
```
kumare3, line 202: `AdminAccess` ->can we just call if `FlyteAdmin`

pkg/controller/nodes/subworkflow/launchplan/admin.go
```diff
@@ -24,6 +24,8 @@ import (
 
 // Executor for Launchplans that executes on a remote FlyteAdmin service (if configured)
 type adminLaunchPlanExecutor struct {
+	Executor
```
kumare3, line 27: I am confused, isnt this the actual struct?

kumare3: or is admin client implementing both methods. If so why we do need to compose of the actual interfaces?

wild-endeavor: I can delete this.

kumare3:
Looks good to me

PR #107, opened 2020-04-11
Backoff also on TooManyRequests and ServerTimeout errors

wild-endeavor:
# TL;DR
We saw that we weren't backing off in all cases where we should have been.  When resource quotas were at their limit for a namespace recently, we still saw the Pod K8s API server endpoint pegged at 8seconds.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
Minor bug, no issue.

### DIFF
--- a/pkg/controller/nodes/task/backoff/handler.go
+++ b/pkg/controller/nodes/task/backoff/handler.go
@@ -145,7 +145,7 @@ func (h *ComputeResourceAwareBackOffHandler) Handle(ctx context.Context, operati
 			return nil
 		}
 
-		if IsResourceQuotaExceeded(err) || apiErrors.IsTooManyRequests(err) {
+		if IsResourceQuotaExceeded(err) || apiErrors.IsTooManyRequests(err) || apiErrors.IsServerTimeout(err) {
 			if !isBlocking {
 				// if the backOffBlocker is not blocking and we are still encountering insufficient resource issue,
 				// we should increase the exponent in the backoff and update the NextEligibleTime
--- a/pkg/controller/nodes/task/backoff/handler.go
+++ b/pkg/controller/nodes/task/backoff/handler.go
@@ -145,22 +145,24 @@ func (h *ComputeResourceAwareBackOffHandler) Handle(ctx context.Context, operati
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/nodes/task/backoff/handler.go
```diff
@@ -145,7 +145,7 @@ func (h *ComputeResourceAwareBackOffHandler) Handle(ctx context.Context, operati
 			return nil
 		}
 
-		if IsResourceQuotaExceeded(err) {
+		if IsResourceQuotaExceeded(err) || apiErrors.IsTooManyRequests(err) {
```
kumare3, line 148: should we add timeout as well?

pkg/controller/nodes/task/backoff/handler.go
```diff
@@ -145,22 +145,24 @@ func (h *ComputeResourceAwareBackOffHandler) Handle(ctx context.Context, operati
...
 		}
 
-		if IsResourceQuotaExceeded(err) {
+		if IsBackOffError(err) {
 			if !isBlocking {
 				// if the backOffBlocker is not blocking and we are still encountering insufficient resource issue,
 				// we should increase the exponent in the backoff and update the NextEligibleTime
 
 				backOffDuration := h.SimpleBackOffBlocker.backOff(ctx)
 				logger.Infof(ctx, "The operation was attempted because the back-off handler is not blocking, but failed due to "+
```
EngHabu, line 154: I think we should print the reason not just "or or or"... unless we dump `err` somewhere else?

pkg/controller/nodes/task/backoff/handler.go
```diff
@@ -145,22 +145,24 @@ func (h *ComputeResourceAwareBackOffHandler) Handle(ctx context.Context, operati
 			return nil
 		}
 
-		if IsResourceQuotaExceeded(err) {
+		if IsBackOffError(err) {
```
EngHabu, line 148: Will this block automatically update the "Ceiling" of the resources? I think that part (updating the ceiling) should only happen on resource quota exceeded error... other errors should continue to exponentially backoff normally

PR #115, opened 2020-04-16
Revert "pass command line args"

wild-endeavor:
Reverts lyft/flytepropeller#63

I believe this PR is causing deploys to fail with
```
Error:
strict mode is on but received keys [map[add_dir_header:false alsologtostderr:false config:/etc/flyte/config*/config.yaml help:false log_backtrace_at::0 log_dir: log_file: log_file_max_size:1800 logtostderr:true skip_headers:false skip_log_headers:false stderrthreshold:2 v:0 vmodule:]] to decode with no config assigned to receive them: failed strict mode check
```

### DIFF
--- a/cmd/controller/cmd/root.go
+++ b/cmd/controller/cmd/root.go
@@ -94,13 +94,11 @@ func init() {
 	rootCmd.AddCommand(viper.GetConfigCommand())
 }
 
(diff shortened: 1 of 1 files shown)


kumare3:
please revert

kumare3:
please revert, but this looks like some wrong config

PR #117, opened 2020-04-19
Improving unit test coverage for simpler cases

kumare3:
# TL;DR
We want to drive overall unit test coverage. Following this checkin we will enforce strict coverage gates

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
NA

## Follow-up issue
NA


kumare3:
@wild-endeavor unit tests are not always for protecting large problems in systems, IMO they serve 2 purposes
1. When you write unit tests you realize if the API (for the function) you have designed, whether it is usable, do you have the right assumptions and sometimes, they serve as a documentation
2. It gives new contributors confidence to change code, knowing that a breakage in the unit test should result in case they make some change that breaks existing assumptions.

PR #119, opened 2020-04-20
Emit User/System permanent error failures for Node

anandswaminathan:
We want to know the workflows that failed through user/system errors.

Unfortunately the way we are setup, this is not easy to achieve. I tried a number of options, and I believe almost everyone of them requires a significant refactor. I have implemented a version to gather opinion.

Alternate options tried:
1) Storing the Execution Error in Nodestatus: This involves replacing `error` with `executionError` - that unfortunately does not bode well in a lot of places, as we use `NodeErrorWithCause` and actually check if `e, ok := err.(*NodeError)`
2) Adding another field `execError` - This could be more confusing, as I do not know when to use which field.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking issue
https://jira.lyft.net/browse/MB-2251

### DIFF
--- a/pkg/controller/nodes/executor.go
+++ b/pkg/controller/nodes/executor.go
@@ -34,18 +34,18 @@ import (
 )
 
 type nodeMetrics struct {
-	Scope                     promutils.Scope
-	FailureDuration           labeled.StopWatch
-	SuccessDuration           labeled.StopWatch
-	UserErrorDuration         labeled.StopWatch
-	SystemErrorDuration       labeled.StopWatch
-	UnknownErrorDuration      labeled.StopWatch
+	Scope                         promutils.Scope
+	FailureDuration               labeled.StopWatch
+	SuccessDuration               labeled.StopWatch
+	UserErrorDuration             labeled.StopWatch
+	SystemErrorDuration           labeled.StopWatch
+	UnknownErrorDuration          labeled.StopWatch
 	PermanentUserErrorDuration    labeled.StopWatch
 	PermanentSystemErrorDuration  labeled.StopWatch
 	PermanentUnknownErrorDuration labeled.StopWatch
-	ResolutionFailure         labeled.Counter
-	InputsWriteFailure        labeled.Counter
-	TimedOutFailure           labeled.Counter
+	ResolutionFailure             labeled.Counter
+	InputsWriteFailure            labeled.Counter
+	TimedOutFailure               labeled.Counter
 
 	InterruptedThresholdHit labeled.Counter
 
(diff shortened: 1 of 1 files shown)


kumare3:
@anandswaminathan can we please associate an issue with this request?

anandswaminathan:
@kumare3 https://jira.lyft.net/browse/MB-2251

pkg/controller/nodes/executor.go
```diff
@@ -353,17 +365,19 @@ func (c *nodeExecutor) handleQueuedOrRunningNode(ctx context.Context, nCtx *node
 		} else {
 			c.metrics.UnknownErrorDuration.Observe(ctx, startTime, endTime)
 		}
+		// When a node fails, we fail the workflow. Independent of number of nodes succeeding/failing, whenever a first node fails,
+		// the entire workflow is failed.
+		if np == v1alpha1.NodePhaseFailing {
```
EngHabu, line 370: In case of retries, do we move from Running -> Failing -> Running -> _Potentially_ Succeeded? or it never hits Failing?

pkg/controller/nodes/executor.go
```diff
@@ -334,6 +337,15 @@ func (c *nodeExecutor) handleQueuedOrRunningNode(ctx context.Context, nCtx *node
 		return executors.NodeStatusUndefined, err
 	}
 
+	if p.GetPhase() == handler.EPhaseUndefined {
```
kumare3, line 340: why can this ever happen, where we get an undefined but no error? That seems like a bug?

pkg/controller/nodes/executor.go
kumare3, line 370: Is this assuming that a node-error will be a workflow error and hence we can count it so?

pkg/controller/nodes/executor.go
anandswaminathan: I don't think it will ever happen. I am not changing the logic.

I believe that check literally means we have a case where phase is not set downstream, and we do not have checks for `handler.EPhaseUndefined` after this.

@surindersinghp @EngHabu if you guys have more context.

pkg/controller/nodes/executor.go
anandswaminathan: @EngHabu Why would we go from Failing to Running.

If we reach failing, we will move to finalize() and then to Failed?

anandswaminathan: ```We can have a case when we never execute any node```
Shouldn't this happen when trying to executing a node, and we should return system error if that happens?

anandswaminathan: Yes

kumare3: there is one problem maybe with this, if we allow run to completion flag, like an issue that was posted. But, I have to think more

anandswaminathan: @EngHabu In case of retries, the phase will be `NodePhaseRetryableFailure`

EngHabu: We can fail before getting to try to run a node... I'm leaning towards what you proposed earlier.. returning a recognizable error, and checking for that at the top level.. I understand it's more work but I think this is an important metric to get right.

anandswaminathan:
@kumare3 These are not workflow stats. You can interpret as workflow if needed. I am merely adding it as a node stat

PR #123, opened 2020-04-23
Error overhaul to isolate User vs system errors

kumare3:
# TL;DR
This change enables every error in Propeller of Plugin to be classfied as a user/system/unknown error and expose it correctly to the higher level. This error is also preserved and can be retrieved from the CRD as well as from the Admin database and enables labeling metrics correctly
TODO: some inner function errors are not handled correctly yet, they may be incorrectly classified as system/unknown errors. (dynamic and branch)

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/300

## Follow-up issue
https://github.com/lyft/flyte/issues/300


pkg/controller/nodes/dynamic/handler.go
```diff
@@ -448,7 +450,7 @@ func (d dynamicNodeTaskNodeHandler) progressDynamicWorkflow(ctx context.Context,
 		}
 
 		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRunning(nil)),
-			handler.DynamicNodeState{Phase: v1alpha1.DynamicNodePhaseFailing, Reason: state.Err.Error()},
+			handler.DynamicNodeState{Phase: v1alpha1.DynamicNodePhaseFailing, Reason: utils.ToString(state.Err)},
```
EngHabu, line 453: I think you will have to preserve the Error here too (not serialize it) because that's what should be recorded as the error on the dynamic node?

pkg/controller/nodes/executor.go
```diff
@@ -449,7 +451,7 @@ func (c *nodeExecutor) handleNode(ctx context.Context, dag executors.DAGStructur
 		}
 		nodeStatus.UpdatePhase(v1alpha1.NodePhaseFailed, v1.Now(), nodeStatus.GetMessage())
 		c.metrics.FailureDuration.Observe(ctx, nodeStatus.GetStartedAt().Time, nodeStatus.GetStoppedAt().Time)
-		return executors.NodeStatusFailed(fmt.Errorf(nodeStatus.GetMessage())), nil
+		return executors.NodeStatusFailed(utils.ParseExecutionError(nodeStatus.GetMessage())), nil
```
EngHabu, line 454: Not a big fan, can we just keep it a strongly typed exception all the way through?

kumare3: ohh i think I will have to implement the DeepCopy?

pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -180,6 +180,8 @@ type NodeStatus struct {
 
 	TaskNodeStatus    *TaskNodeStatus    `json:",omitempty"`
 	DynamicNodeStatus *DynamicNodeStatus `json:"dynamicNodeStatus,omitempty"`
+	// In case of Failing/Failed Phase, an execution error can be optionally associated with the Node
+	Err *ExecutionError `json:"error,omitempty"`
```
EngHabu, line 184: Nit: I would match the name in json and the field... 

```suggestion
	Error *ExecutionError `json:"error,omitempty"`
```

pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -396,6 +398,11 @@ func (in *NodeStatus) UpdatePhase(p NodePhase, occurredAt metav1.Time, reason st
 		n = metav1.Now()
 	}
 
+	if err != nil {
+		in.Err = new(ExecutionError)
```
EngHabu, line 402:
```suggestion
		in.Err = &ExecutionError { 
		                ExecutoinError: err,
		              }
```

pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -396,6 +398,11 @@ func (in *NodeStatus) UpdatePhase(p NodePhase, occurredAt metav1.Time, reason st
...
+		in.Err.ExecutionError = err
```
EngHabu, line 403:
```suggestion
```

pkg/apis/flyteworkflow/v1alpha1/workflow_status.go
```diff
@@ -31,6 +32,9 @@ type WorkflowStatus struct {
 	// the retries the workflow will fail
 	FailedAttempts uint32 `json:"failedAttempts,omitempty"`
 
+	// Stores the Error during the Execution of the Workflow. It is optional and usually associated with Failing/Failed state only
+	Err *ExecutionError `json:"error,omitempty"`
```
EngHabu, line 36: nit:
```suggestion
	Error *ExecutionError `json:"error,omitempty"`
```

pkg/apis/flyteworkflow/v1alpha1/workflow_status.go
```diff
@@ -55,13 +59,25 @@ func (in *WorkflowStatus) UpdatePhase(p WorkflowPhase, msg string) {
 		in.StartedAt = &n
 	}
 
+	if err != nil {
+		in.Err = new(ExecutionError)
```
EngHabu, line 63:
```suggestion
		in.Err = &ExecutionError {
		                  ExecutionError: err,
		              }
```

pkg/apis/flyteworkflow/v1alpha1/workflow_status.go
```diff
@@ -55,13 +59,25 @@ func (in *WorkflowStatus) UpdatePhase(p WorkflowPhase, msg string) {
...
+		in.Err.ExecutionError = err
```
EngHabu, line 64:
```suggestion
```

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -98,7 +98,8 @@ func (d dynamicNodeTaskNodeHandler) handleParentNode(ctx context.Context, prevSt
 func (d dynamicNodeTaskNodeHandler) handleDynamicSubNodes(ctx context.Context, nCtx handler.NodeExecutionContext, prevState handler.DynamicNodeState) (handler.Transition, handler.DynamicNodeState, error) {
 	execContext, dynamicWF, nl, _, err := d.buildContextualDynamicWorkflow(ctx, nCtx)
 	if err != nil {
-		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(
+		// TODO @kumare classify system vs user errors
```
EngHabu, line 101: Another PR?

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -441,42 +442,35 @@ func (d dynamicNodeTaskNodeHandler) progressDynamicWorkflow(ctx context.Context,
...
+	if state.HasFailed() || state.HasTimedOut() {
+		// When the subworkflow either fails or times-out we need to handle failing
 		if dynamicWorkflow.GetOnFailureNode() != nil {
 			// TODO Once we migrate to closure node we need to handle subworkflow using the subworkflow handler
 			logger.Errorf(ctx, "We do not support failure nodes in dynamic workflow today")
 		}
 
-		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRunning(nil)),
-			handler.DynamicNodeState{Phase: v1alpha1.DynamicNodePhaseFailing, Reason: state.Err.Error()},
+		// As we do not support Failure Node, we can just return failure in this case
```
EngHabu, line 452: There was a good reason why we added this extra step in the internal statemachine, I wish I documented that :) let me try to remember, this node requires essentially two rounds to abort, since nodes inside it will first transition to Failing in one round, then their aborts will be called in the next? or maybe that was just an artifact for not having recursive abort? I would rather keep this change in a separate PR though. It seems risky if we get it wrong... and I would like this PR to go in ASAP.

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -441,42 +442,35 @@ func (d dynamicNodeTaskNodeHandler) progressDynamicWorkflow(ctx context.Context,
...
+		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailureErr(state.Err, nil)),
+			handler.DynamicNodeState{Phase: v1alpha1.DynamicNodePhaseFailing, Reason: "Dynamic Workflow execution Failed"},
```
EngHabu, line 454: Do we need to add ExecutionError in the DynamcNodeState to be able to propagate it?

pkg/controller/nodes/subworkflow/launchplan.go
```diff
@@ -98,15 +98,15 @@ func (l *launchPlanHandler) CheckLaunchPlanStatus(ctx context.Context, nCtx hand
...
+		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(core.ExecutionError_USER, errors.RemoteChildWorkflowExecutionFailed, err.Error(), &handler.ExecutionInfo{
 			WorkflowNodeInfo: &handler.WorkflowNodeInfo{LaunchedWorkflowID: childID},
 		})), nil
 	case core.WorkflowExecution_FAILED:
 		errMsg := fmt.Sprintf("launchplan execution failed without explicit error")
 		if wfStatusClosure.GetError() != nil {
 			errMsg = fmt.Sprintf(" errorCode[%s]: %s", wfStatusClosure.GetError().Code, wfStatusClosure.GetError().Message)
 		}
-		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(errors.RemoteChildWorkflowExecutionFailed, errMsg, &handler.ExecutionInfo{
+		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(core.ExecutionError_USER, errors.RemoteChildWorkflowExecutionFailed, errMsg, &handler.ExecutionInfo{
```
EngHabu, line 109: The execution closure error has kind. should we use the kind from there? If it failed due to a system error, should this execution also fail with a system error?

pkg/apis/flyteworkflow/v1alpha1/node_status.go
kumare3: I guess that should work, lower case `error` is a reserved keyword

pkg/controller/nodes/dynamic/handler.go
kumare3: removed

pkg/controller/nodes/executor.go
kumare3: done

pkg/controller/nodes/subworkflow/launchplan.go
kumare3: yes we should

pkg/controller/nodes/dynamic/handler.go
kumare3: I thought about. I dont think so, as the handler is already returning an error

pkg/controller/nodes/dynamic/handler.go
kumare3: ok

pkg/controller/nodes/dynamic/handler.go
kumare3: yes

pkg/controller/nodes/dynamic/handler.go
kumare3: the only thing is, if we want to keep the state running, then i will need to add execution error to dynamic node state

pkg/controller/nodes/dynamic/handler.go
kumare3: yes, added as a tODO to this PR

pkg/controller/executors/node_lookup.go
```diff
@@ -14,24 +14,6 @@ type NodeLookup interface {
 	GetNodeExecutionStatus(ctx context.Context, id v1alpha1.NodeID) v1alpha1.ExecutableNodeStatus
 }
 
-// Implements a de-generate case of NodeLookup, where only one Node is always looked up
```
wild-endeavor, line 17: Why are we removing this?

kumare3: we dont use it, we can add it when we need it

pkg/controller/nodes/dynamic/handler.go
EngHabu: Agreed... let's do that... in the Dynamic-PR

pkg/controller/nodes/subworkflow/launchplan.go
```diff
@@ -44,7 +44,7 @@ func (l *launchPlanHandler) StartLaunchPlan(ctx context.Context, nCtx handler.No
 		if launchplan.IsAlreadyExists(err) {
 			logger.Infof(ctx, "Execution already exists [%s].", childID.Name)
 		} else if launchplan.IsUserError(err) {
-			return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(errors.RuntimeExecutionError, "failed to create unique ID", &handler.ExecutionInfo{
+			return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(core.ExecutionError_USER, errors.RuntimeExecutionError, "failed to create unique ID", &handler.ExecutionInfo{
```
EngHabu, line 47: the error description seems off. no? we should copy the string from `err.String()` probably?

pkg/controller/nodes/subworkflow/launchplan.go
```diff
@@ -69,15 +69,15 @@ func (l *launchPlanHandler) CheckLaunchPlanStatus(ctx context.Context, nCtx hand
...
+		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(core.ExecutionError_SYSTEM, errors.RuntimeExecutionError, "failed to create unique ID", nil)), nil
 	}
 
 	wfStatusClosure, err := l.launchPlan.GetStatus(ctx, childID)
 	if err != nil {
 		if launchplan.IsNotFound(err) { // NotFound
 			errorCode, _ := errors.GetErrorCode(err)
 			err = errors.Wrapf(errorCode, nCtx.NodeID(), err, "launch-plan not found")
-			return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(errorCode, err.Error(), &handler.ExecutionInfo{
+			return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(core.ExecutionError_USER, errorCode, err.Error(), &handler.ExecutionInfo{
```
EngHabu, line 80: hmm is this a user error? if we come here we are saying the system guaranteed we launched something, right? it can't just disappear by any user action... not that we ever hit this error

pkg/controller/nodes/subworkflow/launchplan.go
```diff
@@ -98,15 +98,15 @@ func (l *launchPlanHandler) CheckLaunchPlanStatus(ctx context.Context, nCtx hand
 	case core.WorkflowExecution_ABORTED:
 		wErr = fmt.Errorf("launchplan execution aborted")
 		err = errors.Wrapf(errors.RemoteChildWorkflowExecutionFailed, nCtx.NodeID(), wErr, "launchplan [%s] failed", childID.Name)
```
EngHabu, line 100: Shouldn't this message say "aborted" ?

pkg/controller/nodes/subworkflow/launchplan.go
kumare3: hmm but then this should not happen. It could be a system error where we have a corruption

pkg/controller/nodes/subworkflow/launchplan.go
kumare3: done

PR #126, opened 2020-04-30
Nullable field handling in CRD

kumare3:
# TL;DR
ExecutionError is a nullable field. JSON marshal does not work with null values. This fix addresses this problem

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
NA

## Tracking Issue
https://github.com/lyft/flyte/issues/300
## Follow-up issue
NA

### DIFF
--- a/pkg/apis/flyteworkflow/v1alpha1/node_status.go
+++ b/pkg/apis/flyteworkflow/v1alpha1/node_status.go
@@ -128,7 +128,9 @@ func (in *DynamicNodeStatus) SetDynamicNodePhase(phase DynamicNodePhase) {
 }
 
 func (in *DynamicNodeStatus) SetExecutionError(err *core.ExecutionError) {
-	in.Error = &ExecutionError{ExecutionError: err}
+	if err != nil {
+		in.Error = &ExecutionError{ExecutionError: err}
+	}
 }
 
 func (in *DynamicNodeStatus) Equals(o *DynamicNodeStatus) bool {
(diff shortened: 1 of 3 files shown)


pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -128,7 +128,9 @@ func (in *DynamicNodeStatus) SetDynamicNodePhase(phase DynamicNodePhase) {
 }
 
 func (in *DynamicNodeStatus) SetExecutionError(err *core.ExecutionError) {
-	in.Error = &ExecutionError{ExecutionError: err}
+	if err != nil {
```
EngHabu, line 131: If it's nil, should we set `in.Error = nil` ?

kumare3: hmm interesting point. Like a reset?

pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -128,7 +128,10 @@ func (in *DynamicNodeStatus) SetDynamicNodePhase(phase DynamicNodePhase) {
...
+		in.Error = &ExecutionError{ExecutionError: err}
+	}
+	in.Error = nil
```
EngHabu, line 134: This will reset it all the time... add `else` ?

kumare3: done, good catch

PR #130, opened 2020-05-11
Add project/domain to context for metric labeling

wild-endeavor:
# TL;DR
Add project/domain to context for metric labeling.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Added this line of code
```
ctx = contextutils.WithProjectDomain(ctx, wfDeepCopy.GetExecutionID().Project, wfDeepCopy.GetExecutionID().Domain)
```

## Tracking Issue
NA

## Follow-up issue
NA

### DIFF
--- a/pkg/controller/handler.go
+++ b/pkg/controller/handler.go
@@ -105,7 +105,10 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 	wfDeepCopy := w.DeepCopy()
 	t.Stop()
 	ctx = contextutils.WithWorkflowID(ctx, wfDeepCopy.GetID())
-	ctx = contextutils.WithProjectDomain(ctx, wfDeepCopy.GetExecutionID().Project, wfDeepCopy.GetExecutionID().Domain)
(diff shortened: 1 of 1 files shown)


pkg/controller/handler.go
```diff
@@ -105,6 +105,7 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 	wfDeepCopy := w.DeepCopy()
 	t.Stop()
 	ctx = contextutils.WithWorkflowID(ctx, wfDeepCopy.GetID())
+	ctx = contextutils.WithProjectDomain(ctx, wfDeepCopy.GetExecutionID().Project, wfDeepCopy.GetExecutionID().Domain)
```
EngHabu, line 108: Is there an assumption this will never be nil (for cases when there is no admin) do we have that assumption elsewhere?

wild-endeavor: meh... let's just be safe

PR #141, opened 2020-05-30
Trim error messages read from errors.pb to avoid spaming Etcd. and Admin

EngHabu:
# TL;DR
Containers can generate arbitrary large errors.pb files, this change protects propeller from accidentally loading a large errors.pb, saving it to etcd. and sending it to admin.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/334

### DIFF
--- a/pkg/controller/nodes/task/pre_post_execution.go
+++ b/pkg/controller/nodes/task/pre_post_execution.go
@@ -95,7 +95,9 @@ func (t *Handler) ValidateOutputAndCacheAdd(ctx context.Context, nodeID v1alpha1
 			return nil, err
 		}
 
-		// Errors can be arbitrary long since they are written by
+		// Errors can be arbitrary long since they are written by containers/potentially 3rd party plugins. This ensures
+		// the error message length will never be big enough to cause write failures to Etcd. or spam Admin DB with huge
+		// objects.
 		taskErr.Message = trimErrorMessage(taskErr.Message, t.cfg.MaxErrorMessageLength)
 		return &taskErr, nil
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/nodes/task/transformer.go
```diff
@@ -42,6 +42,14 @@ func ToTaskEventPhase(p pluginCore.Phase) core.TaskExecution_Phase {
 	}
 }
 
+func trimErrorMessage(original string, maxLength int) string {
+	if len(original) <= maxLength {
+		return original
+	}
+
+	return original[0:maxLength/2] + original[len(original)-maxLength/2:]
```
wild-endeavor, line 50: since this is just a string can we do + "\n...\n" +

pkg/controller/nodes/task/pre_post_execution.go
```diff
@@ -94,6 +94,11 @@ func (t *Handler) ValidateOutputAndCacheAdd(ctx context.Context, nodeID v1alpha1
 		if err != nil {
 			return nil, err
 		}
+
+		// Errors can be arbitrary long since they are written by containers/potentially 3rd party plugins. This ensures
+		// the error message length will never be big enough to cause write failures to Etcd. or spam Admin DB with huge
+		// objects.
+		taskErr.Message = trimErrorMessage(taskErr.Message, t.cfg.MaxErrorMessageLength)
```
kumare3, line 101: I think we should trim the error code too? It is a string too right

PR #152, opened 2020-06-10
Implement OnFailurePolicy in Propeller

EngHabu:
# TL;DR
Add support to OnFailurePolicy defined at the Workflow/Subworkflow levels. It defaults to `FAIL_IMMEDIATELY` keeping current behavior the same.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
When set to `FAIL_AFTER_EXECUTABLE_NODES_COMPLETE`, it leads to the workflow continuing to run until all nodes have either succeeded, failed, timedout or skipped.

Screen Shot 2020-06-15 at 1 54 41 PM

## Tracking Issue
https://github.com/lyft/flyte/issues/191


kumare3:
Or description update

pkg/controller/nodes/executor.go
```diff
@@ -602,7 +613,10 @@ func canHandleNode(phase v1alpha1.NodePhase) bool {
 		phase == v1alpha1.NodePhaseSucceeding
 }
 
-func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, execContext executors.ExecutionContext, dag executors.DAGStructure, nl executors.NodeLookup, currentNode v1alpha1.ExecutableNode) (executors.NodeStatus, error) {
+func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, execContext executors.ExecutionContext,
```
kumare3, line 616: I think you will have to change CanExecute, add support for upstream failed state, then just fail the downstream node

pkg/controller/nodes/executor.go
```diff
@@ -535,18 +537,24 @@ func (c *nodeExecutor) handleDownstream(ctx context.Context, execContext executo
...
 		}
-		if state.HasFailed() {
-			logger.Debugf(ctx, "Some downstream node has failed, %s", state.Err)
-			return state, nil
-		}
-		if state.HasTimedOut() {
-			logger.Debugf(ctx, "Some downstream node has timedout")
-			return state, nil
+
+		if state.HasFailed() || state.HasTimedOut() {
```
kumare3, line 546: also similar check in subworkflow?

kumare3:
I am thinking its not ready yet right?

EngHabu:
Ready for review @kumare3

pkg/apis/flyteworkflow/v1alpha1/workflow.go
```diff
@@ -191,6 +191,10 @@ type WorkflowSpec struct {
 	// refer to specific outputs of a subset of the nodes executed in the Workflow. When executing the end-node,
 	// the execution engine will traverse these bindings and assemble the final set of outputs of the workflow.
 	OutputBindings []*Binding `json:"outputBindings,omitempty"`
+
+	// Defines the policy for handling failures whether it's to fail immediately, or let the nodes run
+	// to completion.
+	OnFailurePolicy core.WorkflowMetadata_OnFailurePolicy `json:"onFailurePolicy,omitempty"`
```
kumare3, line 197: should we wrap this in a struct?

kumare3: like
```
FailurePolicy{
enum
... future
}```

This also helps in jsonpb

File: pkg/apis/flyteworkflow/v1alpha1/workflow.go [hunk 4]
```diff
@@ -222,6 +226,10 @@ func (in *WorkflowSpec) FromNode(name NodeID) ([]NodeID, error) {
 	return downstreamNodes, nil
 }
 
+func (in *WorkflowSpec) GetOnFailurePolicy() core.WorkflowMetadata_OnFailurePolicy {
```
kumare3, line 229: we need to use jsonpb to deserialize.

pkg/controller/nodes/executor.go
kumare3: aah you are right

pkg/apis/flyteworkflow/v1alpha1/workflow.go
EngHabu: done

pkg/apis/flyteworkflow/v1alpha1/workflow.go
EngHabu: Do we need jsonpb? this should work, right?

pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -153,6 +153,25 @@ func (b BranchNodePhase) String() string {
 	return "Undefined"
 }
 
+//go:generate enumer --type WorkflowOnFailurePolicy --json --yaml --trimprefix=WorkflowOnFailurePolicy
+
+// Failure Handling Policy
+type WorkflowOnFailurePolicy int32
```
kumare3, line 159: Interesting, so if we add one more element, we will need to add it here?

kumare3: I think I like this, but it does not follow our other enums right?

kumare3:
Looks good to me, I think we need to agree that all top level enums will be handled so.

pkg/apis/flyteworkflow/v1alpha1/iface.go
EngHabu: Is this better?

PR #162, opened 2020-06-24
Redis universal client

EngHabu:
# TL;DR
Add support for Redis Universal Client to enable using different Redis Configurations.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/364


pkg/controller/nodes/task/resourcemanager/config/config.go
```diff
@@ -34,6 +34,9 @@ type Config struct {
 
 // Specific configs for Redis resource manager
 type RedisConfig struct {
+	HostPaths   []string `json:"hostPaths" pflag:",Redis hosts locations."`
+	PrimaryName string   `json:"primaryName" pflag:",Redis primary name."`
```
wild-endeavor, line 38: Can you add a comment for primary name?  what is a primary name?

kumare3: ohh should we just delete the other config attributes then? and use index 0 as the hostpath?

bnsblue: Where can we find the primary name from the redis cluster?

EngHabu: @kumare3  I thought for back compat we shouldn't delete the other one?

@wild-endeavor They call it "MasterName" in redis config but I was trying to avoid using "Master" in our own config...

kumare3: i guess time to support deprecation in config

kumare3:
Is this all the work needed to support Redis universal client?

pkg/controller/nodes/handler/state_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"encoding/base64"
+	"encoding/gob"
+	"testing"
+
+	"github.com/lyft/flytepropeller/pkg/controller/nodes/task/k8s"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDecodeTaskState(t *testing.T) {
```
bnsblue, line 14: Can you add a comment on what this is testing and why/whether that str is meaningful?

bnsblue:
One more q: How does an administrator know what to fill in in the `primaryName` field of the config? Where to look it up?

EngHabu:
Added a comment with a link to the cluster setup.

PR #166, opened 2020-07-23
Add metrics for # of interruptible nodes running

katrogan:
# TL;DR
In order to monitor the roll-out of interruptability it's useful to track the current number of interruptible nodes running in the system.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
_NA_

## Tracking Issue
https://github.com/lyft/flyte/issues/207

## Follow-up issue
_NA_
OR
_https://github.com/lyft/flyte/issues/<number>_


pkg/controller/nodes/executor.go
```diff
@@ -464,6 +468,9 @@ func (c *nodeExecutor) handleNode(ctx context.Context, dag executors.DAGStructur
 		}
 		nodeStatus.UpdatePhase(v1alpha1.NodePhaseFailed, v1.Now(), nodeStatus.GetMessage(), nodeStatus.GetExecutionError())
 		c.metrics.FailureDuration.Observe(ctx, nodeStatus.GetStartedAt().Time, nodeStatus.GetStoppedAt().Time)
+		if nCtx.md.IsInterruptible() {
+			c.metrics.InterruptibleNodes.Dec(ctx)
```
EngHabu, line 472: hmmm I do not know how gauges behave across restarts? will they start from 0 and decrementing them lead to -ve values? from what [I see](https://github.com/prometheus/client_golang/blob/master/prometheus/gauge.go#L107) (just a quick glance I could be wrong), it'll indeed be -ve after a restart. I couldn't find (also another quick glance) docs on whether the server will try to correct that....

long way to say, if we care about this number, we might need to do a bigger change to accurately track it...

katrogan: hm thanks i was thinking of aggregating the value using rawsum but that appears to still be error prone
updated to use two separate counters although my understanding of the node state machine isn't very solid - do we process the transition to queued and from a terminating -> terminated state at most once?

kumare3:
Looks good to me

katrogan:
friendly bump

PR #172, opened 2020-08-06
Remove all dependency on github secret for pull request 

evalsocket:
## Read then delete

# TL;DR
1. Remove github secret dependency from pull request  

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description


evalsocket:
@EngHabu check workflow testing result - https://github.com/evalsocket/flytepropeller/pull/3/checks?check_run_id=952296324

kumare3:
@evalsocket this is really cool, thank you so much, so in flyte we manage all the CI centrally using https://github.com/lyft/boilerplate. We basically modify boilerplate and sync all repos, so that all repos automatically get all the improvements.
What you have done is awesome.

@EngHabu should we directly add @evalsocket's work to boilerplate? Or should we let him merge here and then migrate the work to boilerplate.

LGTM!

.github/workflows/master.yml
```diff
@@ -26,29 +26,13 @@ jobs:
...
-        with:
-          username: "${{ github.actor }}"
-          password: "${{ secrets.GITHUB_TOKEN }}"
-          image_name: ${{ secrets.flytegithub_repo }}/flytepropeller
-          image_tag: latest,${{ github.sha }},${{ needs.bump-version.outputs.version }}
-          push_git_tag: true
-          registry: docker.pkg.github.com
-          build_extra_args: "--compress=true"
-      - uses: engineerd/setup-kind@v0.4.0
+        name: "Checkout"
```
EngHabu, line 29: Let's keep master.yml as is... this is still needed for master merges.

EngHabu: You can do [this trick](https://github.community/t/stop-github-actions-running-on-a-fork/17965/2) to disable running it on master merges on forks

.github/workflows/pull_request.yml
```diff
@@ -8,39 +8,13 @@ jobs:
...
-      - name: Push Docker Image to Dockerhub
-        uses: whoan/docker-build-with-cache-action@v5
-        with:
-          username: "${{ secrets.DOCKERHUB_USERNAME }}"
-          password: "${{ secrets.DOCKERHUB_PASSWORD }}"
-          image_name: ${{ secrets.DOCKERHUB_IMAGE_NAME }}
-          image_tag: ${{ github.sha }}
-          push_git_tag: true
-          build_extra_args: "--compress=true"
+          docker build -t lyft/flytepropeller/flytepropeller:${{ github.sha }} .
```
EngHabu, line 15: Can we continue to use the kind action step? I do like the separation in actions to make it easier to debug failures when they happen...

evalsocket: ok

.github/workflows/master.yml
evalsocket: ok

evalsocket:
@EngHabu all changes are done. I also removed the flytepropeller_repo from master

.github/workflows/master.yml
```diff
@@ -88,4 +86,4 @@ jobs:
         env:
           GO111MODULE: "on"
         with:
-          args: make install && make lint
+          args: make install && make lint
```
EngHabu, line 89: nit
```suggestion
          args: make install && make lint

```

.github/workflows/master.yml
```diff
@@ -40,15 +40,13 @@ jobs:
           build_extra_args: "--compress=true"
       - uses: engineerd/setup-kind@v0.4.0
       - name: End2End
-        env:
-          DOCKER_USERNAME: ${{ github.actor }}
-          DOCKER_PASSWORD: "${{ secrets.GITHUB_TOKEN }}"
         run: |
+          docker build -t lyft/flytepropeller:${{ github.sha }} .
```
EngHabu, line 44: hmmm this is something I would like to dig deeper in. My understanding is that steps will run on the same machine. This means that the previous step that just build the docker image should still have the image on the docker cache. Can we just use it?

.github/workflows/pull_request.yml
```diff
@@ -8,39 +8,18 @@ jobs:
...
-          registry: docker.pkg.github.com
+        name: "Checkout"
+
       - uses: engineerd/setup-kind@v0.4.0
       - name: End2End
-        env:
-          DOCKER_USERNAME: ${{ github.actor }}
-          DOCKER_PASSWORD: "${{ secrets.GITHUB_TOKEN }}"
         run: |
+          docker build -t lyft/flytepropeller:${{ github.sha }} .
```
EngHabu, line 16: Can we continue to use the docker with cache action? we don't need to rebuild the image from scratch every time... specially pulling dependencies.

If my guess is correct, we should still be able to directly `kind load docker-image docker.pkg.github.com/${{PROPELLER}}` because it'll exist in the machine cache... I could be wildly wrong though :)

.github/workflows/master.yml
evalsocket: yes i used it before also and for 100% confirmation i can add imagePullPolicy to `Never`

EngHabu: My point is, can we get rid of this line? because the previous step will ensure this image does exist. right?

Makefile
```diff
@@ -47,3 +50,7 @@ golden:
 .PHONY: generate
 generate: download_tooling
 	@go generate ./...
+
+
+
+
```
EngHabu, line 56: nit:
```suggestion

```

Makefile
```diff
@@ -3,6 +3,9 @@ include boilerplate/lyft/docker_build/Makefile
 include boilerplate/lyft/golang_test_targets/Makefile
 include boilerplate/lyft/end2end/Makefile
 
+PWD := $(shell pwd)
+GIT_HASH := $(shell git log -1 --pretty=format:"%H")
+
```
EngHabu, line 8: Needed?

.github/workflows/pull_request.yml
EngHabu: Is this still needed?

.github/workflows/pull_request.yml
```diff
@@ -8,39 +8,18 @@ jobs:
...
-      - name: Push Docker Image to Github Registry
-        uses: whoan/docker-build-with-cache-action@v5
-        with:
-          username: "${{ github.actor }}"
-          password: "${{ secrets.GITHUB_TOKEN }}"
-          image_name: ${{ secrets.flytegithub_repo }}/flytepropeller
-          image_tag: ${{ github.sha }}
-          push_git_tag: true
-          registry: docker.pkg.github.com
+        name: "Checkout"
```
EngHabu, line 11: Why delete the docker build action? can we continue to use it and just not push the image?

evalsocket:
@EngHabu image is not availbale in secound steps that's why it's failing

.github/workflows/pull_request.yml
```diff
@@ -8,39 +8,27 @@ jobs:
...
       - name: Push Docker Image to Github Registry
         uses: whoan/docker-build-with-cache-action@v5
         with:
           username: "${{ github.actor }}"
           password: "${{ secrets.GITHUB_TOKEN }}"
-          image_name: ${{ secrets.flytegithub_repo }}/flytepropeller
-          image_tag: ${{ github.sha }}
-          push_git_tag: true
+          image_name: lyft/flytepropeller
```
EngHabu, line 17: If you would like to get rid of this action, I would say let's use the github-provided [cache action](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows).

so the action would be something like:
```
- uses: github/cache
   key: "latest"
   path: "/docker_cache"
```

and in the script below (end2end or otherwise), we should follow [this](https://runnable.com/blog/distributing-docker-cache-across-hosts):
```
docker load /docker_cache/latest.tar
docker build...
docker save lyft/flytepropeller:<sha> $(docker history -q lyft/flytepropeller) > /docker_cache/latest.tar
```

If you see some other way to leverage caching please elaborate on how will this work?

EngHabu (approved):
This is awesome! Thank you for going through all the changes. I believe this is exactly what we need. Have you verified the cache is working? Can you make a dummy commit (you can revert it later) to see the cache being hit and used?

.github/workflows/master.yml
```diff
@@ -28,27 +28,24 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: '0'
-      - name: Push Docker Image to Github Registry
-        uses: whoan/docker-build-with-cache-action@v5
+      - id: cache-docker
```
EngHabu, line 31: Any chance we can also push the image to github registry in the master workflow only? They have finally added support for anonymous pull (which was why we had to stick with dockerhub)...

https://docs.github.com/en/packages/getting-started-with-github-container-registry/about-github-container-registry

I'm ok if you want this to go in a separate PR..

evalsocket: no, we don't need that because in the end everyone is using dockerhub registry to pull

evalsocket:
@EngHabu cache is working.

.github/workflows/master.yml
evalsocket: Let's do it in seprate pr

EngHabu:
Looking at the output [here](https://github.com/lyft/flytepropeller/runs/1156137046), I see that it's still building the image from scratch...

EngHabu (approved):

PR #174, opened 2020-08-06
Node execution refactor for Parent-child relationship

anandswaminathan:
Intentions:
The code if rolled out as is - will have no impact or changes to event schema. The code needs to be rolled out and the event version has to be bumped up to V1 for the changes to reflect both in the CRD and the Admin.

Details:
This refactor mainly touches the way node execution status for child nodes are stored in etcd and details relayed to Admin. Following new fields have been introduced : 
- UniqueNodeID - Indicates the unique execution of the node within the workflow. Func(ParentUniqueID, currentNodeId, attempt)
- UniqueParentNodeID - Indicates the unique parent id if node has parent.
- ParentAttempts - Execution attempt of the current node, when the Parent is retried as whole.
- WorkflowMeta - EventVersion: Event Version for a giver workflow. 

In V2,
- For all nodes will start sending uniqueID as node_id in the execution.
- For Dynamic task, the ID of the node execution status has been changed.
- For Branch, no change in the CRD, but events to admin will have uniqueID and parentUniqueID 

TODO : Dirty check needs to fixed. Check with @EngHabu as to how it was added. Any autogen code?

@EngHabu @kumare3 While reviewing, start from node_status.go, and check the logic in branch, subworkflow and dynamic_task. Rest are all just cosmetic.

## Type
 - [X] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/218


pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -550,18 +592,24 @@ func (in *NodeStatus) GetNodeExecutionStatus(ctx context.Context, id NodeID) Exe
...
 	}
+	// By default when a new NodeStatus object is created, we assume that it does not have a parent.
+	// The uniqueID is updated when the parentID is set.
+	uniqueID, err := ComputeUniqueIDForNode(id, "", "")
+	if err != nil {
+		return nil, err
+	}
+	newNodeStatus.UniqueNodeID = &uniqueID
 	newNodeStatus.SetParentTaskID(in.GetParentTaskID())
-	newNodeStatus.SetParentNodeID(in.GetParentNodeID())
```
EngHabu, line 554: I see you removed this... intentionally?

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -38,18 +38,38 @@ func (d dynamicNodeTaskNodeHandler) buildDynamicWorkflowTemplate(ctx context.Con
 	}
 
 	currentAttemptStr := strconv.Itoa(int(nCtx.CurrentAttempt()))
+	uniqueParentID := nCtx.NodeStatus().GetUniqueNodeID()
```
EngHabu, line 41: Something smells here (probably because of the existing code not your new additions)... If you look [here](https://github.com/lyft/flytepropeller/blob/2eae8a1312ce07caafba13eb4bef9a11ca27ebc9/pkg/controller/nodes/dynamic/dynamic_workflow.go#L123). `parentNodeStatus` is this "fake" node we create under the real node, we call it `dynamic-node` or something... so `nCtx.NodeStatus()` is not the same as `parentNodeStatus`... we should decide if we want to continue to have this fake node. I don't think it's required anymore because now we have support for direct node to node relationship... maybe :)

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -38,18 +38,38 @@ func (d dynamicNodeTaskNodeHandler) buildDynamicWorkflowTemplate(ctx context.Con
...
-
+		if nCtx.ExecutionContext().GetEventVersion() == v1alpha1.V0 {
+			subNodeStatus, err = parentNodeStatus.GetNodeExecutionStatus(ctx, newID)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			// In the newer version we create nodestatus for nodes with the same id
+			// A node execution is uniquely identified with the unique id and not the node_id
+			subNodeStatus, err = parentNodeStatus.GetNodeExecutionStatus(ctx, n.Id)
```
EngHabu, line 60: Should we update `GetNodeExecutionStatus` to take in the `currentAttemptStr` as well and return a well formed `subNodeStatus` so you don't have to call `Set*` funcs on it after it's returned? it should have all the information it needs to compute that... right?

pkg/controller/nodes/dynamic/dynamic_workflow.go
anandswaminathan: @EngHabu I have commented there.

We have to until ParentTaskId is removed. That's not happening, until few versions out.

pkg/controller/nodes/dynamic/dynamic_workflow.go
anandswaminathan: True. It's not just attempt. There are several things. But main reason, they don't apply to majority of the cases today. But let's chat about that.

pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -259,7 +259,10 @@ type MutableNodeStatus interface {
 	SetDataDir(DataReference)
 	SetOutputDir(d DataReference)
 	SetParentNodeID(n *NodeID)
+	SetUniqueNodeID(uniqueNodeID *NodeID)
+	SetUniqueParentNodeID(uniqueParentNodeID *NodeID)
 	SetParentTaskID(t *core.TaskExecutionIdentifier)
+	SetParentAttempts(attempt uint32)
```
kumare3, line 265: Hmm What does this mean? How can I child SetParentAttempts?

PR #184, opened 2020-09-15
Implement customizable plugin overrides

katrogan:
# TL;DR
This change handles substituting plugin implementations for a task type besides the default registered for a task type.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1OWYJ2ggr0pWvOdOKkf04NLtj2Rf32dMVYvAYMoyeU48/edit?usp=sharing

## Tracking Issue
https://github.com/lyft/flyte/issues/499

## Follow-up issue
_NA_


katrogan:
friendly ping @kumare3 @EngHabu

pkg/apis/flyteworkflow/v1alpha1/workflow.go
```diff
@@ -298,3 +305,10 @@ type FlyteWorkflowList struct {
 	metav1.ListMeta `json:"metadata"`
 	Items           []FlyteWorkflow `json:"items"`
 }
+
+// This contains workflow-execution specifications and overrides.
+type ExecutionConfig struct {
+	// Maps individual task types to their alternate (non-default) plugin handlers by name.
+	TaskPluginImpls        map[string][]string
+	TaskPluginOverrideMode admin.PluginOverride_MissingPluginBehavior
```
EngHabu, line 313: This is per task type, right? the way it's modeled in admin I mean...

Shouldn't the datastructure be similar here?
```
TaskPluginImpls.  map[string]TaskPluginOverride


type TaskPluginOverride struct {
[]string
missingpluginbehavior
}
```

pkg/controller/nodes/task/handler.go
```diff
@@ -210,7 +216,14 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
 		}
 		for _, tt := range p.RegisteredTaskTypes {
 			logger.Infof(ctx, "Plugin [%s] registered for TaskType [%s]", cp.GetID(), tt)
-			t.plugins[tt] = cp
+			t.defaultPlugins[tt] = cp
```
EngHabu, line 219: I _think_ there is a boolean on `t` to the effect of `IsDefault` we should use that to determine which one is the default for that type

pkg/controller/nodes/task/handler.go
```diff
@@ -230,8 +243,32 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
 	return nil
 }
 
-func (t Handler) ResolvePlugin(ctx context.Context, ttype string) (pluginCore.Plugin, error) {
-	p, ok := t.plugins[ttype]
+func (t Handler) ResolvePlugin(ctx context.Context, ttype string, executionConfig v1alpha1.ExecutionConfig) (pluginCore.Plugin, error) {
+	// If the workflow specifies plugin overrides, check to see if any of the specified plugins for that type are
+	// registered in this deployment of flytepropeller.
+	if len(executionConfig.TaskPluginImpls) > 0 && len(executionConfig.TaskPluginImpls[ttype]) > 0 {
```
EngHabu, line 249: nit: I _think_ the second check is enough but I could be wrong
```suggestion
	if taskPluginImpls := executionConfig.TaskPluginImpls[ttype];len(taskPluginImpls) > 0 {
```

pkg/controller/nodes/task/handler.go
```diff
@@ -230,8 +243,32 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
...
+		if len(t.pluginsForType) > 0 && len(t.pluginsForType[ttype]) > 0 {
+			taskPluginImpls, ok := executionConfig.TaskPluginImpls[ttype]
```
EngHabu, line 251: nit: this is always true (given the condition two lines up), right?
```suggestion
```

pkg/controller/nodes/task/handler.go
```diff
@@ -230,8 +243,32 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
...
+			if ok {
```
EngHabu, line 252:
```suggestion
```

pkg/controller/nodes/task/handler.go
```diff
@@ -230,8 +243,32 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
...
+				pluginsForType := t.pluginsForType[ttype]
+				for _, pluginImplID := range taskPluginImpls {
+					pluginImpl := pluginsForType[pluginImplID]
+					if ok {
+						logger.Debugf(ctx, "Plugin [%s] resolved for Handler type [%s]", pluginImpl.GetID(), ttype)
+						return pluginImpl, nil
+					}
+				}
```
EngHabu, line 260:
```suggestion
```

EngHabu (commented):
Awesome! some nits only

pkg/controller/nodes/task/handler.go
katrogan: oh interesting, the intended behavior then _isn't_ the current one where the last registered plugin for a task type is used? will this break anything if i change it in this PR?

pkg/apis/flyteworkflow/v1alpha1/workflow.go
katrogan: done, thanks @EngHabu

katrogan:
friendly ping @EngHabu

pkg/controller/nodes/task/handler.go
EngHabu: yeah :( it should have flat out thrown an exception (in the current behavior) but looks like that broke after propeller refactor we did last year...

pkg/apis/flyteworkflow/v1alpha1/workflow.go
EngHabu: hmmm Am I looking at an old sha? I don't see this changed...

EngHabu:
You might have local changes that are not pushed?

katrogan:
@EngHabu I don't think so 😅 what do you see missing?

EngHabu:
The default plugins part? using IsDefault to know if it's the default plugin for a particular task type?

pkg/controller/nodes/task/handler.go
katrogan: sorry I missed this reply! It looks like we're using the IsDefault as the global default task/plugin handler: https://github.com/lyft/flytepropeller/blob/master/pkg/controller/nodes/task/handler.go#L215 @EngHabu

pkg/apis/flyteworkflow/v1alpha1/workflow.go
katrogan: https://github.com/lyft/flytepropeller/pull/184/files#diff-79c4ac3e62dcfb7f9b67360a573a9b71R312 ?

pkg/controller/nodes/task/handler.go
EngHabu: Yeah I guess using it for both purposes won't work either. The logic of using the last plugin as the default is very fragile. Depending on the ordering of the imports it can be one plugin or another. I do realize this is not introduced by your PR. Do you mind commenting on this line and filing a ticket to fix it?
We might need a way to configure these in propeller or something... P2 I guess :(

EngHabu (approved):

pkg/controller/nodes/task/handler.go
katrogan: sure, done here: https://github.com/lyft/flyte/issues/516

katrogan:
@EngHabu mind reviewing one more time?

EngHabu (approved):

PR #192, opened 2020-10-01
Define GetProjectLabels in WorkflowMeta

catalinii:
More details https://github.com/lyft/flyte/issues/521


catalinii:
Discussed with @kumare3 and will do just flyteadmin implementation

PR #194, opened 2020-10-08
Add GetMaxAttempts() to taskExecutionMetadata

catalinii:
https://github.com/lyft/flyte/issues/469

### DIFF
--- a/pkg/controller/nodes/task/taskexec_context.go
+++ b/pkg/controller/nodes/task/taskexec_context.go
@@ -26,6 +26,7 @@ var (
 )
 
 const IDMaxLength = 50
+const DefaultMaxAttempts = 1
 
 type taskExecutionID struct {
 	execName string
@@ -42,8 +43,9 @@ func (te taskExecutionID) GetGeneratedName() string {
 
 type taskExecutionMetadata struct {
 	handler.NodeExecutionMetadata
-	taskExecID taskExecutionID
-	o          pluginCore.TaskOverrides
+	taskExecID  taskExecutionID
+	o           pluginCore.TaskOverrides
+	maxAttempts uint32
 }
 
 func (t taskExecutionMetadata) GetTaskExecutionID() pluginCore.TaskExecutionID {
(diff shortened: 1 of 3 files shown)


pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -140,13 +145,15 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 	}
 
 	resourceNamespacePrefix := pluginCore.ResourceNamespace(t.resourceManager.GetID()).CreateSubNamespace(pluginCore.ResourceNamespace(pluginID))
+	maxAttempts := nCtx.Node().GetRetryStrategy().MinAttempts
```
anandswaminathan, line 148: @catalinii This field a bit tricky. Just see how it is used. The tricky part comes in interpretation of this field when you want to attempt 'x' times more or total 'x'

catalinii: It is total maxAttempts according to the code.

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -140,13 +145,18 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 	}
 
 	resourceNamespacePrefix := pluginCore.ResourceNamespace(t.resourceManager.GetID()).CreateSubNamespace(pluginCore.ResourceNamespace(pluginID))
+	maxAttempts := uint32(0)
```
anandswaminathan, line 148: I don't think this is valid. Attempt has to be greater than 0 right?

@katrogan @kumare3 Do we set default maxAttempts to 3 ?

katrogan: yeah it looks like we always set the min attempts to 1 + the user specified value: https://github.com/lyft/flytepropeller/blob/96c9ee9e2a9288101de846be8f49fce8c244ea2c/pkg/compiler/transformers/k8s/utils.go#L24

anandswaminathan (dismissed):

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -140,13 +145,18 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 	}
 
 	resourceNamespacePrefix := pluginCore.ResourceNamespace(t.resourceManager.GetID()).CreateSubNamespace(pluginCore.ResourceNamespace(pluginID))
+	maxAttempts := uint32(3)
```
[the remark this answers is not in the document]
catalinii: Done

anandswaminathan (approved):

PR #199, opened 2020-10-19
Introduce default plugin for task type

katrogan:
# TL;DR
Introduce default plugin for task type configurable in the flytepropeller config.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/516

## Follow-up issue
_NA_


go.mod
```diff
@@ -23,11 +22,10 @@ require (
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/lyft/datacatalog v0.2.1
 	github.com/lyft/flyteidl v0.18.9
-	github.com/lyft/flyteplugins v0.5.12
+	github.com/lyft/flyteplugins v0.5.14-0.20201019203656-dbcbf808fc35
```
katrogan, line 25: won't merge until this is an official release

pkg/controller/nodes/task/handler_test.go
```diff
@@ -150,41 +151,59 @@ func Test_task_Setup(t *testing.T) {
...
 			pluginIDs: map[pluginCore.TaskType]string{k8sPluginType: k8sPluginType},
 		}, false},
-		{"no-default", testPluginRegistry{
-			core: []pluginCore.PluginEntry{corePluginEntry}, k8s: []pluginK8s.PluginEntry{k8sPluginEntry},
+		{"no-default", testPluginRegistry{}, map[string]config.EnabledPlugins{
+			corePluginType: {DefaultPluginTasks: []string{corePluginType}},
+			k8sPluginType:  {DefaultPluginTasks: []string{k8sPluginType}},
 		}, wantFields{
-			pluginIDs: map[pluginCore.TaskType]string{corePluginType: corePluginType, k8sPluginType: k8sPluginType},
+			pluginIDs: map[pluginCore.TaskType]string{},
```
katrogan, line 179: @EngHabu can you let me know if this change is correct? we shouldn't expect any plugins right if the registry is empty?

EngHabu: 👍

pkg/controller/nodes/task/config/config.go
```diff
@@ -45,23 +44,38 @@ type BarrierConfig struct {
 	CacheTTL  config.Duration `json:"cache-ttl" pflag:", Max duration that a barrier would be respected if the process is not restarted. This should account for time required to store the record into persistent storage (across multiple rounds."`
 }
 
+type EnabledPlugins struct {
```
EngHabu, line 47: nit:
```suggestion
type PluginConfig struct {
```

pkg/controller/nodes/task/config/config.go
```diff
@@ -45,23 +44,38 @@ type BarrierConfig struct {
...
+	DefaultPluginTasks []string `json:"default-plugins-tasks" pflag:",Tasks for which this plugin is the default implementation"`
+}
+
 type TaskPluginConfig struct {
-	EnabledPlugins []string `json:"enabled-plugins" pflag:",Plugins enabled currently"`
+	EnabledPlugins map[string]EnabledPlugins `json:"enabled-plugins" pflag:",Plugins enabled currently"`
```
EngHabu, line 52: This is a "breaking" change in as far as Propeller/deployment is concerned... just a note to make sure you bump the version, and make sure the PR & the release are prefixed with "Breaking"... you can, of course, choose to leave this and add a new field and deprecate this one later.... your choice

pkg/controller/nodes/task/config/config.go
```diff
@@ -45,23 +44,38 @@ type BarrierConfig struct {
 	CacheTTL  config.Duration `json:"cache-ttl" pflag:", Max duration that a barrier would be respected if the process is not restarted. This should account for time required to store the record into persistent storage (across multiple rounds."`
 }
 
+type EnabledPlugins struct {
+	DefaultPluginTasks []string `json:"default-plugins-tasks" pflag:",Tasks for which this plugin is the default implementation"`
```
EngHabu, line 48: nit:
```suggestion
	DefaultForTaskTypes []string `json:"default-for-task-types" pflag:",Task types for which this plugin is the default handler."`
```

pkg/controller/nodes/task/config/config.go
```diff
@@ -45,23 +44,38 @@ type BarrierConfig struct {
...
 	MaxDuration config.Duration `json:"max-duration" pflag:",The cap of the backoff duration"`
 }
 
-func (p TaskPluginConfig) GetEnabledPluginsSet() sets.String {
-	s := sets.NewString()
-	for _, e := range p.EnabledPlugins {
-		cleanedPluginName := strings.Trim(e, " ")
-		cleanedPluginName = strings.ToLower(cleanedPluginName)
-		s.Insert(cleanedPluginName)
+func cleanString(source string) string {
```
EngHabu, line 60: Thank you!

pkg/controller/nodes/task/handler.go
```diff
@@ -215,10 +215,19 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
...
+		println(fmt.Sprintf("for plugin [%s], registered task types: [%+v] and default task types [%+v]",
+			p.ID, p.RegisteredTaskTypes, p.DefaultForTaskTypes))
 		for _, tt := range p.RegisteredTaskTypes {
-			logger.Infof(ctx, "Plugin [%s] registered for TaskType [%s]", cp.GetID(), tt)
-			// TODO(katrogan): Make the default task plugin assignment more explicit (https://github.com/lyft/flyte/issues/516)
-			t.defaultPlugins[tt] = cp
+			for _, defaultTaskType := range p.DefaultForTaskTypes {
+				if defaultTaskType == tt {
+					if existingHandler, alreadyDefaulted := t.defaultPlugins[tt]; alreadyDefaulted {
+						logger.Warnf(ctx, "TaskType [%s] has multiple default handlers specified: [%s] and [%s]",
```
EngHabu, line 224: I think this should fail starting up Propeller.

katrogan: actually I might leave it as a warning for the allPluginsEnabled case since there's no clear order of operations there.

katrogan:
friendly ping @EngHabu

EngHabu (commented):
I think we are close!
I'll try to write down my understand and plz correct me where I got it wrong.
We have these different settings:
1- Plugin Author decides which task types a plugin can handle,
2- Plugin Author decides which task types a plugin can be the default for (We should consider punting on this)
3- Admin decides which plugins are enabled
4- Admin decides which plugins are defaults for which task types

Some validation before attempting the resolution:
- If a plugin exist in #4 but the plugin isn't enabled (not in #3 nor allpluginsEnabled) or doesn't exist, fail?

The resolution for which defaults are should be:
- If a task type exist in #4, that plugin choice wins
- Otherwise, tasks defined in #2 take precedence, if there is a conflict, fail

I think we should consider getting rid of #2 and just consider #1 as such... In what scenarios would these be different?

katrogan:
> We have these different settings:
> 1- Plugin Author decides which task types a plugin can handle,
> 2- Plugin Author decides which task types a plugin can be the default for (We should consider punting on this)
> 3- Admin decides which plugins are enabled
> 4- Admin decides which plugins are defaults for which task types
> 
> Some validation before attempting the resolution:
> 
> If a plugin exist in #4 but the plugin isn't enabled (not in #3 nor allpluginsEnabled) or doesn't exist, fail?
> The resolution for which defaults are should be:
> 
> If a task type exist in #4, that plugin choice wins
> Otherwise, tasks defined in #2 take precedence, if there is a conflict, fail
> I think we should consider getting rid of #2 and just consider #1 as such... In what scenarios would these be different?

2 is not supported. Plugin authors cannot specify defaults, only registered tasks. Plugin resolution in propeller should ignore any default values plugin authors attempt to set.

re: 3 this will require an idl and admin change. currently admin only provides overrides for specific plugins for a task type. if we want the net affect of this change to be 3, such that the defaults are specified in admin rather than in the propeller config I can close this PR and take that approach.

pkg/controller/nodes/task/handler.go
```diff
@@ -216,16 +220,30 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
 			return regErrors.Wrapf(err, "failed to load plugin - %s", p.ID)
 		}
 		for _, tt := range p.RegisteredTaskTypes {
-			logger.Infof(ctx, "Plugin [%s] registered for TaskType [%s]", cp.GetID(), tt)
-			// TODO(katrogan): Make the default task plugin assignment more explicit (https://github.com/lyft/flyte/issues/516)
-			t.defaultPlugins[tt] = cp
+			for _, defaultTaskType := range p.DefaultForTaskTypes {
```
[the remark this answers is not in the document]
katrogan: so the value exists in the struct defined in plugins, but it is overwritten by propeller when reading from the config. i defined it in the plugin struct because that is what we're using to schlep around plugins in this resolution code but it's not meant to be assigned there. if it's preferable, i can create a separate struct in propeller, copy all the fields over and remove the DefaultForTaskTypes from the plugins struct

katrogan:
re 2: see why it's not supported currently here: https://github.com/lyft/flytepropeller/pull/199#discussion_r515361774 (one of your comments was on an outdated file version, so not sure if that led to any confusion)

i updated to error when a default for a task type is specified for a plugin that is not enabled. other than that the order of default preference should indeed be 4, then 1

pkg/controller/nodes/task/config/config.go
```diff
@@ -46,22 +50,63 @@ type BarrierConfig struct {
...
+	DefaultForTaskTypes []string
+}
+
+func cleanString(source string) string {
+	cleaned := strings.Trim(source, " ")
+	cleaned = strings.ToLower(cleaned)
+	return cleaned
+}
+
+func (p TaskPluginConfig) GetEnabledPlugins() (map[string]PluginConfig, error) {
```
EngHabu, line 73:
```suggestion
type PluginID = string
type TaskType = string
func (p TaskPluginConfig) GetEnabledPlugins() (map[PluginID]PluginConfig, error) {
```

pkg/controller/nodes/task/plugin_config.go
```diff
@@ -8,31 +8,35 @@ import (
...
+	logger.Infof(ctx, "Enabled plugins: %v", pluginsConfigMeta.EnabledPlugins.List())
 	logger.Infof(ctx, "Loading core Plugins, plugin configuration [all plugins enabled: %v]", allPluginsEnabled)
 	for _, cpe := range pr.GetCorePlugins() {
 		id := strings.ToLower(cpe.ID)
-		if !allPluginsEnabled && !enabledPlugins.Has(id) {
+		if !allPluginsEnabled && !pluginsConfigMeta.EnabledPlugins.Has(id) {
 			logger.Infof(ctx, "Plugin [%s] is DISABLED (not found in enabled plugins list).", id)
 		} else {
 			logger.Infof(ctx, "Plugin [%s] ENABLED", id)
+			cpe.DefaultForTaskTypes = pluginsConfigMeta.AllDefaultForTaskTypes[id]
```
EngHabu, line 39: I think this will throw if `cfg == nil`.. no?

pkg/controller/nodes/task/plugin_config.go
```diff
@@ -8,31 +8,35 @@ import (
...
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/lyft/flytepropeller/pkg/controller/nodes/task/config"
 	"github.com/lyft/flytepropeller/pkg/controller/nodes/task/k8s"
 )
 
 func WranglePluginsAndGenerateFinalList(ctx context.Context, cfg *config.TaskPluginConfig, pr PluginRegistryIface) ([]core.PluginEntry, error) {
 	allPluginsEnabled := false
-	enabledPlugins := sets.NewString()
+	pluginsConfigMeta := config.PluginsConfigMeta{}
```
EngHabu, line 18: Should this be:
```suggestion
	pluginsConfigMeta := config.PluginsConfigMeta{
	    AllDefaultForTaskTypes: map[PluginID][]TaskType{}
	}
```
To avoid a nil ref below in case `cfg == nil` ?

pkg/controller/nodes/task/handler.go
```diff
@@ -204,28 +204,49 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
 		return err
 	}
 
+	// Not every task type will have a default plugin specified in the flytepropeller config.
+	// That's fine, we resort to using the plugins' static RegisteredTaskTypes as a fallback further below.
+	fallbackTaskHandlerMap := make(map[string]map[string]pluginCore.Plugin)
```
EngHabu, line 209: nit: use those alias types for the `strings` here to clarify what they refer to

pkg/controller/nodes/task/handler.go
```diff
@@ -234,6 +255,20 @@ func (t *Handler) Setup(ctx context.Context, sCtx handler.SetupContext) error {
...
 	}
 
+	// Read from the fallback task handler map for any remaining tasks without a defaultPlugins registered handler.
+	for taskType, registeredPlugins := range fallbackTaskHandlerMap {
+		if _, ok := t.defaultPlugins[taskType]; ok {
+			break
+		}
+		if len(registeredPlugins) != 1 {
+			logger.Errorf(ctx, "Multiple plugins registered to handle task type: %s. ([%+v])", taskType, registeredPlugins)
+			return regErrors.New(fmt.Sprintf("Multiple plugins registered to handle task type: %s. ([%+v])", taskType, registeredPlugins))
```
EngHabu, line 265: nit:
```suggestion
			return regErrors.New(fmt.Sprintf("Multiple plugins registered to handle task type: %s. ([%+v]). Use default-for-task-type config option to choose the desired plugin.", taskType, registeredPlugins))
```

EngHabu (approved):
I think that's it! Thank you a ton for walking me through the code the other day.
Just a couple of tiny nits

pkg/controller/nodes/task/plugin_config.go
katrogan: updated

pkg/controller/nodes/task/handler.go
katrogan: done

EngHabu (approved):

PR #205, opened 2020-11-19
Fix branch compiler and execution

EngHabu:
# TL;DR
Fixes branch compiler and branch execution.

## The compiler had two wrong assumptions:
- All nodes in a branch nodes will have the same inputs and those inputs must be bound in the parent node. This is an unnecessary restriction. Removed.
- Nodes inside a branch nodes will have globally unique IDs. This is not necessarily true. Added code to build unique node ids.

## Execution runtime had a regression:
A prior PR removed the persistence of datadir in the workflow CRD. Branch handler relies on that persistence to find the selected branch's outputs location. The fix is to reverse that setting and update the selected branch's output location to the parent's... This way once the execution of the branch is done, the outputs are already available in the parent node's location.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/602


pkg/apis/flyteworkflow/v1alpha1/workflow.go
```diff
@@ -157,7 +157,7 @@ func (in *Inputs) UnmarshalJSON(b []byte) error {
 
 func (in *Inputs) MarshalJSON() ([]byte, error) {
 	if in == nil {
-		return []byte{}, nil
+		return json.Marshal(nil)
```
kumare3, line 160: i guess we could make this a constant and re-use it?

pkg/compiler/validators/utils.go
```diff
@@ -115,6 +117,26 @@ func withVariableName(param *core.Variable) (newParam *core.Variable, ok bool) {
 	return
 }
 
+func JoinVariableMapsUniqueKeys(m1, m2 map[string]*core.Variable) (map[string]*core.Variable, error) {
```
kumare3, line 120: do you imply union of maps?

kumare3:
This looks like a big change, a lot of null checks and fixes.

pkg/compiler/validators/branch.go
```diff
@@ -31,55 +31,63 @@ func validateBranchInterface(w c.WorkflowBuilder, node c.NodeBuilder, errs error
...
+		if otherCase.ThenNode == nil {
 			errs.Collect(errors.NewValueRequiredErr(node.GetId(), "IfElse.Case.ThenNode"))
 			continue
 		}
 
-		n := w.NewNodeBuilder(block.ThenNode)
+		cases = append(cases, otherCase.ThenNode)
+	}
+
+	if elseNode := node.GetBranchNode().IfElse.GetElseNode(); elseNode != nil {
```
kumare3, line 59: you missed out on the else.Fail case

kumare3:
Looks good to me, except for one comment. To be honest, its very hard to verify the entire logic :)

pkg/compiler/validators/node.go
```diff
@@ -81,6 +86,8 @@ func ValidateBranchNode(w c.WorkflowBuilder, n c.NodeBuilder, requireParamType b
 	if elseNode := n.GetBranchNode().IfElse.GetElseNode(); elseNode != nil {
 		wrapperNode := w.NewNodeBuilder(elseNode)
 		subNodes = append(subNodes, wrapperNode)
+	} else if defaultError := n.GetBranchNode().IfElse.GetError(); defaultError == nil {
```
kumare3: do you not handle the  error today?

EngHabu: nope...

kumare3 (approved):

PR #210, opened 2020-12-03
[Compiler] Compiling a compiled WF leads to compile errors

EngHabu:
# TL;DR
When a workflow has two nodes both of type WorkflowNodes and attempt to execute the same SubWF, the compiler attempts to compile the SubWorkflow twice (opportunity for optimization).. When it does the second time, it fails because validating Start and End nodes doesn't pass due to them not having targets (by design)... the fix is to handle start and end nodes specially...
The ultimate fix is to avoid compiling twice to begin with.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/633

### DIFF
--- a/pkg/compiler/validators/node.go
+++ b/pkg/compiler/validators/node.go
@@ -120,6 +120,10 @@ func ValidateNode(w c.WorkflowBuilder, n c.NodeBuilder, validateConditionTypes b
 		errs.Collect(errors.NewValueRequiredErr("<node>", "Id"))
 	}
 
+	if n.GetId() == c.StartNodeID || n.GetId() == c.EndNodeID {
+		return true
+	}
+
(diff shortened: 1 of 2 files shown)


wild-endeavor (approved):
but why doesn't it fail the first time?  Do they have targets the first time?

EngHabu:
The first time they (start and end nodes) don't even exist, we add them during the compilation phase... so the second time, they look like regular nodes we try to validate...

kumare3:
@EngHabu I am ok with this now, but I feel this is very counter-intuitive. We start with the assumption that we dont have start and end nodes, but then we check for them

kumare3:
Why can we not check for the subworkflow id in a map?

kumare3 (approved):

PR #212, opened 2020-12-21
Option for cleanup pod after last retry failure

tnsetting:
# TL;DR
The default behaviour after the last retry failure is leaving the pod and it will be cleaned when gc happens. In that case, before gc happens the pods still consume k8s resource and after a few executions of a miss configure workflow, no new execution could be done on the namespace because of quota. 

When enabling the option `cleanup-last-retry`, when last retry failed, the pod will also be cleanup. 

When last retry failed the node will still go to `EPhaseRetryableFailure` and in the `handleRetryableFailure` function, the pod will be cleaned up and node transit to failed state.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/651

## Follow-up issue
_NA_
OR
_https://github.com/lyft/flyte/issues/<number>_

### DIFF
--- a/pkg/controller/nodes/executor_test.go
+++ b/pkg/controller/nodes/executor_test.go
@@ -479,7 +479,7 @@ func TestNodeExecutor_RecursiveNodeHandler_Recurse(t *testing.T) {
 		}
 
 		var err *v1alpha1.ExecutionError
-		if p == v1alpha1.NodePhaseFailing || p == v1alpha1.NodePhaseFailed {
+		if p == v1alpha1.NodePhaseFailing || p == v1alpha1.NodePhaseFailed || p == v1alpha1.NodePhaseRetryableFailure{
 			err = &v1alpha1.ExecutionError{ExecutionError: &core.ExecutionError{Code: "test", Message: "test"}}
 		}
 		ns := &v1alpha1.NodeStatus{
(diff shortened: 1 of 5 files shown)


tnsetting:
@kumare3 This is the PR I mentioned earlier. Please take a look.

kumare3:
@tnsetting can you create an issue and refer it in the description?

pkg/controller/nodes/executor_test.go
```diff
@@ -963,6 +963,69 @@ func TestNodeExecutor_RecursiveNodeHandler_Recurse(t *testing.T) {
...
+		adminClient := launchplan.NewFailFastLaunchPlanExecutor()
+		execIface, err := NewExecutor(ctx, config.GetConfig().NodeConfig, store, enQWf, mockEventSink, adminClient,
+			adminClient, 10, "s3://bucket", fakeKubeClient, catalogClient, promutils.NewTestScope())
+		assert.NoError(t, err)
+		exec := execIface.(*nodeExecutor)
+		exec.cleanupLastRetry = true
+		exec.nodeHandlerFactory = hf
+
+		h := &nodeHandlerMocks.Node{}
+		h.On("Handle",
```
kumare3, line 980: you can start using the new interface for writing mock handlers, i.e., in this case
```
h.OnHandle(....).Return(...)
```
It becomes simpler to refactor

pkg/controller/nodes/executor_test.go
```diff
@@ -963,6 +963,69 @@ func TestNodeExecutor_RecursiveNodeHandler_Recurse(t *testing.T) {
...
+			mock.MatchedBy(func(ctx context.Context) bool { return true }),
+			mock.MatchedBy(func(o handler.NodeExecutionContext) bool { return true }),
+		).Return(handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRetryableFailure(core.ExecutionError_USER, "x", "y", nil)), nil)
+		h.On("FinalizeRequired").Return(true)
+		h.On("Finalize", mock.Anything, mock.Anything).Return(nil)
+		h.On("Abort", mock.Anything, mock.Anything, mock.Anything).Return(nil)
```
kumare3, line 1017: you may want to add a case where abort fails?
`.OnAbort(...).Return(fmt.errorf(..))`

kumare3:
Looks great, just couple small comments.

honnix:
@kumare3 Do you think this makes sense to be the default behaviour instead of staying behind a flag? As I user I would expect things being cleaned up after a series of failed retries.

tnsetting:
Close the PR since we might have some problem with logs uploading if the pods are kill too soon. So we will use the GC (max ttl to 1 hour) for now.

PR #216, opened 2021-01-13
[Compiler] Add execution edges only for top level nodes

EngHabu:
# TL;DR
Limits the addition of execution edges (upstream/downstream dependencies) to top level nodes. The main issue arouse with branch nodes as follows:
Before this PR:
- The compiler would traverse all nodes (including branch nodes) to validate them and their execution units (tasks, wfs, lps.. etc.)
- Branch nodes come with their branches embedded into them.
- The compiler pulls those nodes up and just references them.
- At the end of the validation, the compiler adds downstream dependencies for the start node to all nodes that have no upstream dependencies. This makes sure Propeller traverses and executes these nodes.
- At the end of the validation, the compiler adds upstream dependencies for the end node from all nodes that have no downstream dependencies, this ensures that propeller doesn't execute the end-node until all nodes in the workflow finish executing.
- Because branches in branch nodes are flattened out, and they are leaf nodes (have no downstream dependencies), the compiler ends up adding execution links from them to the end-node.
- At runtime, propeller ends up choosing one and only one branch, meaning all other branches are marked as skipped, this makes the end node un-executable because one of its upstream dependencies was skipped.
- The sub-workflow succeeds though (no failures, no more nodes to execute) but does not produce outputs.
- If there are subsequent dependencies to such a workflow, they will fail complaining about the lack of outputs produced.

With this PR:
- The compiler will only add execution edges for top level nodes. branches in branch nodes will no longer have edges from/to start/end nodes.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/667


pkg/compiler/builders.go
```diff
@@ -22,9 +22,10 @@ type workflowBuilder struct {
 	// These are references to all subgraphs and tasks passed to CompileWorkflow. They will be passed around but will
 	// not show in their entirety in the final Graph. The required subset of these will be added to each subgraph as
 	// the compile traverses them.
-	allLaunchPlans  map[string]c.InterfaceProvider
-	allTasks        c.TaskIndex
-	allSubWorkflows c.WorkflowIndex
+	allLaunchPlans          map[string]c.InterfaceProvider
+	allTasks                c.TaskIndex
+	allSubWorkflows         c.WorkflowIndex
+	allCompiledSubWorkflows c.WorkflowIndex
```
kumare3, line 28: interesting, this looks like a big change ha?

kumare3:
so now we believe the compiler is the most complicated piece in the system. to be honest I am losing track of the compiler. I think its going to get even more complicated in the future - we probably need to document it better and eventually might need a refactor pass.
But before that, lets try to capture all weird cases and have then as regression test suite

pkg/compiler/builders.go
EngHabu: Not quite, it used to keep the non-compiled workflows in allSubworkflows and when they get compiled, we replace them in-place. Now they live in two separate lists so we know which ones we compiled before...

EngHabu:
I agree. It probably deserves a rewrite :-) not just a refactor. I've been adding test cases whenever we catch issues. 
Filed [this](https://github.com/lyft/flyte/issues/670) to add a regression test suite... 
Keep in mind that branches were never tested end to end... and they do add an interesting layer of complexity...

pkg/compiler/test/compiler_test.go
```diff
@@ -101,18 +101,96 @@ func marshalProto(t *testing.T, filename string, p proto.Message) {
...
+
+func TestBranches(t *testing.T) {
+	errors.SetConfig(errors.Config{IncludeSource: true})
+	assert.NoError(t, filepath.Walk("testdata/branch", func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+
+		t.Run(path, func(t *testing.T) {
+			if !strings.HasSuffix(path, "success_1.json") {
```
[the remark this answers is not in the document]
EngHabu: 👍

wild-endeavor (approved):

pkg/compiler/builders.go
kumare3: aah ok, this is better :)

pkg/compiler/workflow_compiler.go
```diff
@@ -230,12 +236,23 @@ func (w workflowBuilder) ValidateWorkflow(fg *flyteWorkflow, errs errors.Compile
 			continue
 		}
 
-		if _, foundUpStream := wf.upstreamNodes[nodeID]; !foundUpStream {
-			wf.AddExecutionEdge(c.StartNodeID, nodeID)
+		// Nodes that do not have a upstream dependencies means they do not rely on the workflow inputs to execute.
+		// This is a rare but possible occurrence, by explicitly adding an execution edge from the start node to these
+		// nodes, we ensure that propeller starts executing the workflow by running all such nodes and then their
+		// downstream dependencies.
+		if topLevelNodes.Has(nodeID) {
```
kumare3, line 243: this would have caused all kinds of problems for non input graphs?

kumare3 (approved):

pkg/compiler/workflow_compiler.go
EngHabu: Notice the change doesn't start adding downstream/upstream dependencies... that was already there... the only addition/change is making sure we only add them for top level nodes....

wild-endeavor (approved):

bnsblue (approved):
Approve for merging

PR #219, opened 2021-01-31
Introducing FlytePropeller TURBO Mode #minor

kumare3:
# TL;DR
This PR adds a new configuration to FlytePropeller which is enabled by default. This improves the transition latency of FlytePropeller by an order of magnitude or more. Absolute comparison to come soon.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently FlytePropeller event loop is designed to make the smallest forward progress in each round. There are cases in which the round progress is purely book-keeping. For example moving a workflow from accepted to Running state. Moving a node from new to queued to running state. This is done so that state is resilient from crashes.

the turbo mode, essentially waits for the one of these rounds to complete and optimistically progresses to the next round using the updated state. To add safety to the optimistic progress we can configure bounds.

The psuedo algorithm
```
w -> Original workflow state
w1 -> updated workflow state

while w1 != w and round < configured:
     w1 = try_mutate(w)
     if w1 != w:
         record(w1)
         w = w1
         round++
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/676

### DIFF
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -47,8 +47,8 @@ var (
 			MaxNodeRetriesOnSystemFailures: 3,
 			InterruptibleFailureThreshold:  1,
 		},
-		EnableFastFollow: true,
-		MaxStreakLength:  5,
+		EnableTurboMode: true,
+		MaxStreakLength: 5,
 	}
 )
 
@@ -76,8 +76,8 @@ type Config struct {
 	MaxDatasetSizeBytes    int64                `json:"max-output-size-bytes" pflag:",Maximum size of outputs per task"`
 	KubeConfig             KubeClientConfig     `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
 	NodeConfig             NodeConfig           `json:"node-config,omitempty" pflag:",config for a workflow node"`
-	EnableFastFollow       bool                 `json:"enable-fast-follow" pflag:",Boolean flag that enables Fast Follow mode, this makes Propeller proceed to another round on successful write to etcD."`
-	MaxStreakLength        int                  `json:"max-streak-length" pflag:",Maximum number of consecutive rounds that one propeller worker can use for one workflow if fast follow mode is enabled."`
+	EnableTurboMode        bool                 `json:"enable-turbo-mode" pflag:",Boolean flag that enables Turbo-mode, this makes Propeller proceed to another round on successful write to etcD."`
+	MaxStreakLength        int                  `json:"max-streak-length" pflag:",Maximum number of consecutive rounds that one propeller worker can use for one workflow if turbo-mode is enabled."`
 }
 
 type KubeClientConfig struct {
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -74,6 +74,7 @@ type Config struct {
 	MaxDatasetSizeBytes    int64                `json:"max-output-size-bytes" pflag:",Maximum size of outputs per task"`
 	KubeConfig             KubeClientConfig     `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
 	NodeConfig             NodeConfig           `json:"node-config,omitempty" pflag:",config for a workflow node"`
+	EnableFastFollow       bool					`json:"enable-fast-follow" pflag:",Boolean flag that enables Fast Follow mode, this makes Propeller proceed to another round on successful write to etcD."`
 }
 
 type KubeClientConfig struct {
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -47,8 +47,7 @@ var (
 			MaxNodeRetriesOnSystemFailures: 3,
 			InterruptibleFailureThreshold:  1,
 		},
-		EnableTurboMode: true,
-		MaxStreakLength: 5,
+		MaxStreakLength: 5, // Turbo mode is enabled by default
 	}
 )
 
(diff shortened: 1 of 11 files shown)


pkg/controller/config/config.go
```diff
@@ -74,6 +74,7 @@ type Config struct {
 	MaxDatasetSizeBytes    int64                `json:"max-output-size-bytes" pflag:",Maximum size of outputs per task"`
 	KubeConfig             KubeClientConfig     `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
 	NodeConfig             NodeConfig           `json:"node-config,omitempty" pflag:",config for a workflow node"`
+	EnableFastFollow       bool					`json:"enable-fast-follow" pflag:",Boolean flag that enables Fast Follow mode, this makes Propeller proceed to another round on successful write to etcD."`
```
EngHabu, line 77: nit: formatting seems to have more spaces... tabs?

pkg/controller/handler.go
```diff
@@ -181,39 +181,53 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
-			// If the end result is a terminated workflow, we remove the labels
-			// We add a completed label so that we can avoid polling for this workflow
-			SetCompletedLabel(mutatedWf, time.Now())
-			ResetFinalizers(mutatedWf)
+		// TODO we will need to call updatestatus when it is supported. But to preserve metadata like (label/finalizer) we will need to use update
+
+		// update the GetExecutionStatus block of the FlyteWorkflow resource. UpdateStatus will not
+		// allow changes to the Spec of the resource, which is ideal for ensuring
+		// nothing other than resource status has been updated.
+		newWf, updateErr := p.wfStore.Update(ctx, mutatedWf, workflowstore.PriorityClassCritical)
```
EngHabu, line 213: Time to switch to UpdateStatus?

EngHabu (dismissed):
Sweet!

pkg/controller/handler.go
kumare3: I want to do that as a follow up. It is a little dangerous to do that at the moment, because we add a finalizer in some places. Do not want to de-stabilize

pkg/controller/config/config.go
kumare3: will fix formatting

pkg/controller/handler.go
```diff
@@ -180,40 +186,62 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
-			if mutatedWf.Status.Equals(&w.Status) {
-				logger.Info(ctx, "WF hasn't been updated in this round.")
-				return nil
+	maxLength := p.cfg.MaxStreakLength
+	if maxLength <= 0 {
+		maxLength = 1
+	}
+
+	for streak = 0; streak < maxLength; streak++ {
+		mutatedWf, err := p.TryMutateWorkflow(ctx, w)
```
EngHabu, line 198: Do we measure round time inside TryMuttateWorkflow? won't this be a bit misleading? should we measure it here and record it `n` times instead?

pkg/controller/config/config.go
```diff
@@ -74,6 +76,8 @@ type Config struct {
 	MaxDatasetSizeBytes    int64                `json:"max-output-size-bytes" pflag:",Maximum size of outputs per task"`
 	KubeConfig             KubeClientConfig     `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
 	NodeConfig             NodeConfig           `json:"node-config,omitempty" pflag:",config for a workflow node"`
+	EnableTurboMode        bool                 `json:"enable-turbo-mode" pflag:",Boolean flag that enables Turbo-mode, this makes Propeller proceed to another round on successful write to etcD."`
```
EngHabu, line 79: Should we really have two flags? or just have it enabled all the time and use `MaxStreakLength <= 1` meaning it essentially has no effect?

pkg/controller/workflowstore/inmemory.go
```diff
@@ -50,8 +50,10 @@ func (i *InmemoryWorkflowStore) UpdateStatus(ctx context.Context, w *v1alpha1.Fl
 		if w.Name != "" && w.Namespace != "" {
 			if m, ok := i.store[w.Namespace]; ok {
 				if _, ok := m[w.Name]; ok {
-					m[w.Name] = w
-					return w, nil
+					newW := w.DeepCopy()
+					newW.ResourceVersion = w.ResourceVersion + "x"
```
EngHabu, line 54: What's this?

kumare3: every successful write changes the resource version for local testing

pkg/controller/config/config.go
kumare3: that is interesting

pkg/controller/handler.go
kumare3: ya, you are right, we should add it to TryMutateWorkflow

pkg/controller/handler.go
```diff
@@ -180,40 +184,66 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
+	for streak = 0; streak < maxLength; streak++ {
+		t := p.metrics.RoundTime.Start(ctx)
+		mutatedWf, err := p.TryMutateWorkflow(ctx, w)
+		if err != nil {
+			// NOTE We are overriding the deepcopy here, as we are essentially ingnoring all mutations
+			// We only want to increase failed attempts and discard any other partial changes to the CRD.
+			mutatedWf = RecordSystemError(w, err)
+			p.metrics.SystemError.Inc(ctx)
+		} else if mutatedWf == nil {
+			return nil
```
EngHabu, line 204: Is this a legit case? that looks wrong... I know it existed... maybe at least add a log line here?

pkg/controller/handler.go
```diff
@@ -180,40 +184,66 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
-			SetCompletedLabel(mutatedWf, time.Now())
-			ResetFinalizers(mutatedWf)
+		// TODO we will need to call updatestatus when it is supported. But to preserve metadata like (label/finalizer) we will need to use update
+
+		// update the GetExecutionStatus block of the FlyteWorkflow resource. UpdateStatus will not
+		// allow changes to the Spec of the resource, which is ideal for ensuring
+		// nothing other than resource status has been updated.
+		newWf, updateErr := p.wfStore.Update(ctx, mutatedWf, workflowstore.PriorityClassCritical)
+		if updateErr != nil {
+			return updateErr
```
EngHabu, line 228: Do you intentionally only want to measure the latency on success?

pkg/controller/workflowstore/inmemory.go
EngHabu: Mind adding a comment to "NewInMemoryWorkflowStore" to expect this behavior?

EngHabu (commented):
I left a couple of comments to write a bit more code docs...

pkg/controller/workflowstore/inmemory.go
kumare3: sure

pkg/controller/handler.go
kumare3: nope, adding a defer is not okayed by the linter as it thinks it is a resource leak

EngHabu: I agree with the linter...

EngHabu (approved):

PR #225, opened 2021-03-02
Add support for custom KubeClient

regadas:
This is an initial take on adding support for a custom KubeClient supplied by plugins. If `KubeClient` is added for a remote cluster we need to ignore `OwnerReferences`

@kumare3 @EngHabu let me know what you think. 

This will require https://github.com/flyteorg/flyteplugins/pull/154 and then a `flyteplugins` version bump

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/<number>

### DIFF
--- a/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -97,22 +97,24 @@ type PluginManager struct {
 	kubeClient      pluginsCore.KubeClient
 	metrics         PluginMetrics
 	// Per namespace-resource
-	backOffController     *backoff.Controller
-	resourceLevelMonitor  *ResourceLevelMonitor
-	ignoreOwnerReferences bool
+	backOffController             *backoff.Controller
+	resourceLevelMonitor          *ResourceLevelMonitor
+	overrideInjectOwnerReferences *bool
+	overrideInjectFinalizer       *bool
 }
 
-func (e *PluginManager) AddObjectMetadata(taskCtx pluginsCore.TaskExecutionMetadata, o k8s.Resource, cfg *config.K8sPluginConfig) {
+func (e *PluginManager) AddObjectMetadata(taskCtx pluginsCore.TaskExecutionMetadata, o client.Object, cfg *config.K8sPluginConfig) {
 	o.SetNamespace(taskCtx.GetNamespace())
 	o.SetAnnotations(utils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), utils.CopyMap(taskCtx.GetAnnotations())))
 	o.SetLabels(utils.UnionMaps(o.GetLabels(), utils.CopyMap(taskCtx.GetLabels()), cfg.DefaultLabels))
 	o.SetName(taskCtx.GetTaskExecutionID().GetGeneratedName())
 
-	if !e.ignoreOwnerReferences {
+	if e.overrideInjectOwnerReferences != nil && *e.overrideInjectOwnerReferences {
 		o.SetOwnerReferences([]metav1.OwnerReference{taskCtx.GetOwnerReference()})
 	}
 
-	if cfg.InjectFinalizer {
+	overrideInjectFinalizer := e.overrideInjectFinalizer
+	if (overrideInjectFinalizer != nil && *overrideInjectFinalizer) || cfg.InjectFinalizer {
 		f := append(o.GetFinalizers(), finalizer)
 		o.SetFinalizers(f)
 	}
@@ -398,8 +400,8 @@ func NewPluginManager(ctx context.Context, iCtx pluginsCore.SetupContext, entry 
 	}
 
 	kubeClient := iCtx.KubeClient()
-	if entry.NewKubeClient != nil {
-		kc, err := entry.NewKubeClient(ctx)
+	if entry.CustomKubeClient != nil {
+		kc, err := entry.CustomKubeClient(ctx)
 		if err != nil {
 			return nil, err
 		}
@@ -419,7 +421,7 @@ func NewPluginManager(ctx context.Context, iCtx pluginsCore.SetupContext, entry 
 	}
 
 	workflowParentPredicate := func(o metav1.Object) bool {
-		if entry.IgnoreOwnerReferences {
+		if entry.OverrideInjectOwnerReferences != nil && *entry.OverrideInjectOwnerReferences {
 			return true
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -107,8 +95,25 @@ type PluginManager struct {
 	kubeClient      pluginsCore.KubeClient
 	metrics         PluginMetrics
 	// Per namespace-resource
-	backOffController    *backoff.Controller
-	resourceLevelMonitor *ResourceLevelMonitor
+	backOffController     *backoff.Controller
+	resourceLevelMonitor  *ResourceLevelMonitor
+	ignoreOwnerReferences bool
```
kumare3, line 100: how about finalizer? Having a finalizer is fine is it?

regadas: Hi @kumare3 sorry for the delay. Yea, I just double-checked this and it's better if we have an option to override the k8s config `inject-finalizer` property.

Updated the PR with it.

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -107,8 +95,27 @@ type PluginManager struct {
...
```
EngHabu, line 100: Can we rename these two variables to:
```
overrideInjectOwnerReference *bool
overrideInjectFinalizer *bool
```

regadas: 👍 sounds good!

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -399,14 +418,18 @@ func NewPluginManager(ctx context.Context, iCtx pluginsCore.SetupContext, entry
 		Type: entry.ResourceToWatch,
 	}
 
-	ownerKind := iCtx.OwnerKind()
 	workflowParentPredicate := func(o metav1.Object) bool {
+		if entry.OverrideInjectOwnerReferences != nil && *entry.OverrideInjectOwnerReferences {
```
kumare3, line 422: can we not add a label? I guess but as step 1 this is fine. This may be too chatty?

kumare3:
Looks good to me, can you please rebase and also, fix the linter?

regadas:
Includes changes from #238

pkg/controller/nodes/task/k8s/plugin_manager.go
regadas: Talked IRL. When I think the idea would be to target this in a follow-up PR? right @kumare3?

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -109,8 +97,27 @@ type PluginManager struct {
 	kubeClient      pluginsCore.KubeClient
 	metrics         PluginMetrics
 	// Per namespace-resource
-	backOffController    *backoff.Controller
-	resourceLevelMonitor *ResourceLevelMonitor
+	backOffController             *backoff.Controller
+	resourceLevelMonitor          *ResourceLevelMonitor
+	overrideInjectOwnerReferences *bool
```
kumare3, line 102: so the term override here is confusing, as it does not say what is the behavior?
so true here indicates we are overriding? if so the condition below is wrong right, if this is not specified then, we will inject owner references, if this is set and is true then we will and if this is set and is false we will skip?

kumare3: a better term might be - disableOwnerReferenceInjection *bool? Not sure

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -109,8 +97,27 @@ type PluginManager struct {
...
+	o.SetAnnotations(utils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), utils.CopyMap(taskCtx.GetAnnotations())))
+	o.SetLabels(utils.UnionMaps(o.GetLabels(), utils.CopyMap(taskCtx.GetLabels()), cfg.DefaultLabels))
+	o.SetName(taskCtx.GetTaskExecutionID().GetGeneratedName())
+
+	if e.overrideInjectOwnerReferences != nil && *e.overrideInjectOwnerReferences {
+		o.SetOwnerReferences([]metav1.OwnerReference{taskCtx.GetOwnerReference()})
+	}
+
+	overrideInjectFinalizer := e.overrideInjectFinalizer
+	if (overrideInjectFinalizer != nil && *overrideInjectFinalizer) || cfg.InjectFinalizer {
```
kumare3, line 117: this condition again, is a little confusing
we will also inject if 
cfg.InjectFinalizer is true?
OR
if the override sets it to true?

should it not be
overrideInjectFinalizer != nil && *overrideInjectFinalizer && cfg.InjectFinalizer

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -109,8 +97,26 @@ type PluginManager struct {
...
+	disableInjectFinalizer       bool
+}
+
+func (e *PluginManager) AddObjectMetadata(taskCtx pluginsCore.TaskExecutionMetadata, o client.Object, cfg *config.K8sPluginConfig) {
+	o.SetNamespace(taskCtx.GetNamespace())
+	o.SetAnnotations(utils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), utils.CopyMap(taskCtx.GetAnnotations())))
+	o.SetLabels(utils.UnionMaps(o.GetLabels(), utils.CopyMap(taskCtx.GetLabels()), cfg.DefaultLabels))
+	o.SetName(taskCtx.GetTaskExecutionID().GetGeneratedName())
+
+	if !e.disableInjectOwnerReferences {
```
kumare3, line 112: so much cleaner

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -64,6 +64,14 @@ func (t taskExecutionMetadata) GetMaxAttempts() uint32 {
 	return t.maxAttempts
 }
 
+func (t taskExecutionMetadata) GetSecurityContext() core.SecurityContext {
```
kumare3, line 67: so @anandswaminathan is working on plumbing this all the way through. But I guess we can keep this, when Anand rebases, he can delete it

kumare3 (dismissed):

pkg/controller/nodes/task/taskexec_context.go
regadas: sounds good!

kumare3 (approved):

PR #226, opened 2021-03-03
Pass command-line arguments 

evalsocket:
# TL;DR
Pass command-line arguments

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/167


kumare3:
I am not sure if we should have strict-mode false - @EngHabu ?

evalsocket:
@EngHabu any update on this PR ?

EngHabu (dismissed):

cmd/controller/cmd/root.go
```diff
@@ -97,11 +97,12 @@ func init() {
 	rootCmd.AddCommand(viper.GetConfigCommand())
 }
 
-func initConfig(_ *cobra.Command, _ []string) error {
+func initConfig(cmd *cobra.Command, _ []string) error {
 	configAccessor = viper.NewAccessor(config.Options{
-		StrictMode:  true,
+		StrictMode:  false,
```
EngHabu, line 102: I don't think this is the complete fix here... I understand it'll allow it to pass... I think the more fundamental fix is to make InitializePFlags keep track of all the flags (and persistent flags) that it sees at the time of initialization. Once it does, it let those flags be skipped while everything else must be recognized (if strictMode is true)...

kumare3: @EngHabu is the reason for the problem here that, there are other modules registering flags, and or config, will not recognize them?

evalsocket:
@EngHabu Any update on this?....you can test it with my image `evalsocket/flytepropeller:latest`

PR #229, opened 2021-03-06
Migrate references from Lyft -> Flyteorg, golang 1.16, new kubeapi/client-go, cleanup dependencies #minor

kumare3:
# TL;DR
This is a housekeeping PR, but has the potential to improve a lot of small issues in flytepropeller
 - Migrates all references to use flyteorg instead of lyft
 - Removes un-used and dead dependencies
 - Removes dependency on Lyft's fork of K8s API
 - Uses latest client golang and latest controller-runtime
 - Use latest version of golang - 1.16
 - Generate some missing code for CRD parameters
 
## Type
 - [ ] Bug Fix
 - [x] Feature / Housekeeping
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/806


kumare3:
cc @regadas this is the propeller PR, that you might want to rebase from

katrogan (approved):

pkg/controller/executors/kube.go
```diff
@@ -41,14 +44,34 @@ func (c fallbackClientReader) List(ctx context.Context, list runtime.Object, opt
...
 // calls if it failed. Write calls will always go to raw client directly.
-func NewFallbackClient(cachedClient, rawClient client.Client) client.Client {
-	return client.DelegatingClient{
-		Reader: fallbackClientReader{
-			orderedClients: []client.Client{cachedClient, rawClient},
-		},
-		StatusClient: rawClient,
-		Writer:       rawClient,
-	}
+func NewFallbackClientBuilder() *FallbackClientBuilder {
```
regadas, line 75: I think we could make `FallbackClientBuilder` private and return here `cluster.ClientBuilder`.

The main reason being that we are not adding yet additional attrs to it and we lose the type after `WithUncached` calls.

EngHabu: As a general rule of thumb in go (unlike java), one should ["accept interfaces and return structs"](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a). This allows the receiver to accept the return in whatever interface they define..

EngHabu (approved):

PR #232, opened 2021-03-11
Limit parallelism of task nodes

kumare3:

# TL;DR
This PR proposes an improvement in FlytePropeller throughput, by introducing bounds on the maximum latency of a  flyte propeller round. This can improve throughput by potentially increasing the number of rounds per second and number of free workers.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
FlytePropeller performs a modified DFS algorithm. The algorithm does not terminate, once some quota has been met, In every round, it cycles through the entire graph. This can affect the round performance. Lets call this refresh rate.
Problems with reduced refresh rate / higher latency are as follows,
- Jitter: hence transition rate reduces. 
- If some nodes cause Kube API  server throttling, which is bound to happen because some other nodes ate up the resources, then multiple rounds exacerbate the problem
- Throughput of FlytePropeller drops significantly
- Aborts are complicated, as partial workflow node successes are recorded and FlyteAdmin returns an error,  if trying to abort a succeeded node - `TODO this should be a separate fix in flyte admin's statemachine, ignore two terminal phase transitions`
- In case of complex `state regeneration logic` the round latency increases further. This further affects the refresh rate and thus greatly impacts transition latency

**Solution**
The solution proposed here essentially reduces the thrash by, capping the maximum number of `task` nodes that can be traversed per round. This solves the problem mentioned previously, but does have a problem that may affect some cases
1. In case of theoretically infinite resources and no contention, we are artificially reducing the number of nodes that can run in parallel. Users will now have to override this value per execution.
2. This is not adaptive, so if you have multiple executions using some parallelism limit, does not guarantee the requested parallelism, this is because you may have constrain on the resources themselves. 
But in all cases the problem is not exacerbated.

As a follow up, we may want to add some tasks plugins to request exception from parallelism limits. This is because they may be using an abundant resource like an API call.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/476

## Follow-up issue
TODO: Implement Flytekit control for users to specifiy parallelism within a workflow.


pkg/controller/nodes/executor.go
```diff
@@ -1,3 +1,19 @@
+// Core Nodes Executor implementation
```
katrogan, line 1: Thank you for adding this!

pkg/controller/nodes/executor.go
```diff
@@ -659,11 +676,32 @@ func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, execContext exe
...
+		if currentNode.GetKind() == v1alpha1.NodeKindTask && nodeStatus.GetPhase() == v1alpha1.NodePhaseQueued {
+			maxParallelism := execContext.GetExecutionConfig().MaxParallelism
+			if maxParallelism > 0 {
+				// If we are queued, let us see if we can proceed within the node parallelism bounds
+				if execContext.CurrentParallelism() >= maxParallelism {
+					logger.Infof(ctx, "Maximum Parallelism for task nodes achieved [%d] >= Max [%d], Round will be short-circuited.", execContext.CurrentParallelism(), maxParallelism)
+					return executors.NodeStatusRunning, nil
+				}
+				// We know that Propeller goes through each workflow in a single thread, thus every node is really processed
+				// sequentially. So, we can continue - now that we know we are under the parallelism limits and increment the
```
katrogan, line 694: where do you increment the parallelism?

pkg/controller/nodes/executor.go
```diff
@@ -659,11 +676,32 @@ func (c *nodeExecutor) RecursiveNodeHandler(ctx context.Context, execContext exe
...
 
+		// Now if the node is of type task, then let us check if we are within the parallelism limit, only if the node
+		// has been queued already
+		if currentNode.GetKind() == v1alpha1.NodeKindTask && nodeStatus.GetPhase() == v1alpha1.NodePhaseQueued {
+			maxParallelism := execContext.GetExecutionConfig().MaxParallelism
+			if maxParallelism > 0 {
+				// If we are queued, let us see if we can proceed within the node parallelism bounds
+				if execContext.CurrentParallelism() >= maxParallelism {
+					logger.Infof(ctx, "Maximum Parallelism for task nodes achieved [%d] >= Max [%d], Round will be short-circuited.", execContext.CurrentParallelism(), maxParallelism)
+					return executors.NodeStatusRunning, nil
```
katrogan, line 691: why do you proceed to running if the current parallelism **exceeds** the max?

katrogan (dismissed):

pkg/controller/nodes/executor.go
kumare3: the parallelism is incremented only when the task is running in the task handler

pkg/controller/nodes/executor.go
kumare3: we dont proceed to running, this is the nodeStatus :(, this implies the node is running so check it again. But the task - start is skipped

katrogan: ah okay sorry i got confused by what proceed meant in the comment above

katrogan (approved):

EngHabu (dismissed):
sweet!

EngHabu (approved):

PR #240, opened 2021-03-19
Large workflows will not cause workflows to stay in running state #minor

kumare3:

# TL;DR
- Workflows will be moved to Failing state in case they are very large and etcD rejects them
- Incase of errors, the whole round will be retried, until system errors are hit and then eventually the workflow will be given up
- Large Workflows here refer to workflows with large number of unique tasks -> usually > 1k, Nodes > 2.5k (or even more). the limitation is from etcD. If you use map tasks (array tasks), that is not a problem as the state is handled through dynamic compression.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/363


pkg/controller/handler.go
```diff
@@ -227,6 +237,21 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 		newWf, updateErr := p.wfStore.Update(ctx, mutatedWf, workflowstore.PriorityClassCritical)
 		if updateErr != nil {
 			t.Stop()
+			if workflowstore.IsWorkflowTooLarge(updateErr) {
+				logger.Errorf(ctx, "Failed storing workflow to the store, reason: %s", updateErr)
+				// Workflow is too large, we will mark the workflow as failed and
```
EngHabu, line 242: and ?

pkg/controller/handler.go
```diff
@@ -227,6 +237,21 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 		newWf, updateErr := p.wfStore.Update(ctx, mutatedWf, workflowstore.PriorityClassCritical)
 		if updateErr != nil {
 			t.Stop()
+			if workflowstore.IsWorkflowTooLarge(updateErr) {
```
EngHabu, line 240: Should we move this logic to [here](https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/handler.go#L201-L202) since it's where we already check for update failures, ignore mutated workflows, increment counters... etc. ?

pkg/controller/handler.go
kumare3: no and

pkg/controller/handler.go
kumare3: we cannot, as the workflow is too large error comes from workflowstore

kumare3: but good point, i need to increment the counter

kumare3: Also I am not sure if we need to handle this as system error and increment it

EngHabu (dismissed):

EngHabu (approved):

PR #242, opened 2021-03-19
Pod Mutating Webhook & Secret Annotation Injector

EngHabu:

# TL;DR
Add a Pod Mutating WebHook cmd for flytepropeller. The current setup only implements a secrets injector that can either inject secrets available to FlytePropeller or k8s secrets.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
* `flytepropeller webhook init-certs` command issues a CA/Cert/PrivateKey and store them into a secret given a secret name
* `flytepropeller webhook` creates a MutatingWebhookConfigration object in etcd., a webhook server and registers it with API Server. From there:
    1. It starts listening to Pods labeled with `inject-flyte-secrets: true`. 
    2. Parses out annotations on the pod to know which secrets to inject
    3. Lookups the env (envVars and mounted files) for the secret, if found it'll inject it directly into the pod
    4. Adds k8s EnvFrom or VolumeMountSource to mount the right secret into the Pod
* Plugin Manager change to automatically add `inject-flyte-secrets: true` to CRDs if the task has secrets in its TaskTemplate.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/800


cmd/controller/cmd/webhook.go
```diff
@@ -0,0 +1,158 @@
+package cmd
```
kumare3, line 1: package documentation?

pkg/controller/nodes/task/secretmanager/secrets.go
```diff
@@ -38,6 +41,41 @@ func (f FileEnvSecretManager) Get(ctx context.Context, key string) (string, erro
 	return string(b), err
 }
 
+func (f FileEnvSecretManager) GetForSecret(ctx context.Context, secret *coreIdl.Secret) (string, error) {
```
kumare3, line 44: documentation, that explains the logic?

pkg/controller/nodes/task/secretmanager/secrets.go
```diff
@@ -38,6 +41,41 @@ func (f FileEnvSecretManager) Get(ctx context.Context, key string) (string, erro
...
+	lookup := strings.ToUpper(secret.Key)
```
kumare3, line 45: key is not required parameter anymore, group is
so this needs to be inverted

pkg/controller/nodes/task/secretmanager/secrets.go
```diff
@@ -38,6 +41,41 @@ func (f FileEnvSecretManager) Get(ctx context.Context, key string) (string, erro
...
+
+	envVar := fmt.Sprintf("%s%s", f.envPrefix, lookup)
+	v, ok := os.LookupEnv(envVar)
+	if ok {
+		logger.Debugf(ctx, "Secret found %s", v)
+		return v, nil
+	}
+
+	fileLookup := secret.Key
+	if len(secret.Group) > 0 {
```
kumare3, line 58: samething

pkg/utils/secrets/marshaler.go
```diff
@@ -0,0 +1,87 @@
...
+	decodedRaw, err := encoding.DecodeString(strings.ToUpper(encoded))
+	if err != nil {
+		return encoded, err
+	}
+
+	return string(decodedRaw), nil
+}
+
+func MarshalSecretsToMapStrings(secrets []*core.Secret) (map[string]string, error) {
+	res := make(map[string]string, len(secrets))
```
kumare3, line 38: Why? why not keep it protobuf?

kumare3: what if you just serializeit to protobuf string and then retrieved it from the string representation?

pkg/webhook/global_secrets.go
```diff
@@ -0,0 +1,71 @@
+package webhook
```
kumare3, line 1: documentation

pkg/webhook/k8s_secrets.go
```diff
@@ -0,0 +1,112 @@
+package webhook
```
kumare3, line 1: docs

pkg/webhook/global_secrets.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (g GlobalSecrets) Inject(ctx context.Context, secret *coreIdl.Secret, p *corev1.Pod) (newP *corev1.Pod, injected bool, err error) {
+	v, err := g.envSecretManager.GetForSecret(ctx, secret)
+	if err != nil {
+		return p, false, err
+	}
+
+	switch secret.MountRequirement {
+	case coreIdl.Secret_FILE:
+		return nil, false, fmt.Errorf("cannot use FILE requirement with global secret [%v/%v]", secret.Group, secret.Key)
```
kumare3, line 35:
```suggestion
		return nil, false, fmt.Errorf("global secrets can only be injected as environment variables [%v/%v]", secret.Group, secret.Key)
```

pkg/webhook/k8s_secrets.go
```diff
@@ -0,0 +1,112 @@
...
+			volumeName = secret.Group + EnvVarGroupKeySeparator + secret.Key
+			p.Spec.Volumes = append(p.Spec.Volumes, corev1.Volume{
+				Name: volumeName,
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: secret.Group,
+						Items: []corev1.KeyToPath{
+							{
+								Key:  secret.Key,
+								Path: secret.Key,
```
kumare3, line 45: the path should be a combination of the group and the key right?

pkg/webhook/k8s_secrets.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (i K8sSecretInjector) Inject(ctx context.Context, secret *core.Secret, p *corev1.Pod) (newP *corev1.Pod, injected bool, err error) {
+	switch secret.MountRequirement {
+	case core.Secret_ANY:
+		fallthrough
+	case core.Secret_FILE:
+		volumeName := secret.Key
+		if len(secret.Group) > 0 && len(secret.Key) > 0 {
+			volumeName = secret.Group + EnvVarGroupKeySeparator + secret.Key
+			p.Spec.Volumes = append(p.Spec.Volumes, corev1.Volume{
```
kumare3, line 37: how about making this a separate helper method?

pkg/webhook/pod.go
```diff
@@ -0,0 +1,197 @@
+package webhook
```
kumare3, line 1: docs

pkg/webhook/pod.go
```diff
@@ -0,0 +1,197 @@
...
+	return generateMutatePath(pod.GroupVersionKind())
+}
+
+func generateMutatePath(gvk schema.GroupVersionKind) string {
+	return "/mutate-" + strings.Replace(gvk.Group, ".", "-", -1) + "-" +
+		gvk.Version + "-" + strings.ToLower(gvk.Kind)
+}
+
+func (pm PodMutator) CreateMutationWebhookConfiguration(namespace string) (*admissionregistrationv1.MutatingWebhookConfiguration, error) {
+	caBuff, err := ReadFile(filepath.Join(pm.cfg.CertDir, "ca.crt"))
```
kumare3, line 134: why cant we use ioutil.ReadFile? buffer?

kumare3:
few comments

cmd/controller/cmd/webhook.go
EngHabu: I added docs to each command... I don't know what would make sense to put in the commands package docs... please review

pkg/utils/secrets/marshaler.go
EngHabu: yeah... moved to that in latest version

pkg/webhook/k8s_secrets.go
EngHabu: No this is always a "subpath" of wherever you mount this volume to in each container.

pkg/webhook/pod.go
EngHabu: yeah... I was using it elsewhere that needed a buffer... but not anymore... good call!

cmd/controller/cmd/webhook.go
```diff
@@ -0,0 +1,206 @@
...
+	// Creates a MutationConfig to instruct ApiServer to call this service whenever a Pod is being created.
+	err = createMutationConfig(ctx, kubeClient, secretsWebhook)
+	if err != nil {
+		return err
+	}
+
+	mgr, err := manager.New(kubecfg, manager.Options{
+		Port:          cfg.ListenPort,
+		CertDir:       cfg.CertDir,
+		Namespace:     limitNamespace,
```
kumare3, line 137: if we add multiple namespaces - how does this work?

pkg/utils/secrets/marshaler.go
kumare3: I dont follow, why is this code still here then?

pkg/utils/secrets/marshaler.go
```diff
@@ -0,0 +1,86 @@
...
+	for index, s := range secrets {
+		if _, found := core.Secret_MountType_name[int32(s.MountRequirement)]; !found {
+			return nil, fmt.Errorf("invalid mount requirement [%v]", s.MountRequirement)
+		}
+
+		encodedSecret := marshalSecret(s)
+		res[annotationPrefix+strconv.Itoa(index)] = encodedSecret
+
+		if len(encodedSecret) > totalAnnotationSizeLimitB {
+			return nil, fmt.Errorf("secret descriptor cannot exceet [%v]", totalAnnotationSizeLimitB)
```
kumare3, line 65:
```suggestion
			return nil, fmt.Errorf("secret descriptor cannot exceed [%v]", totalAnnotationSizeLimitB)
```

pkg/utils/secrets/marshaler.go
kumare3: I guess the question is you are keeping every secret still a separate thing in the annotations is it? instead of just making it into one key/value?

kumare3 (approved):
lgtm, can you update the readme for flytepropeller to add that it contains this too?

kumare3 (approved):

PR #251, opened 2021-04-02
#patch Bug Fixes - resiliency in Dynamic spec loading failures and Workqueue metrics name fix

kumare3:

## Read then delete
- Make sure to use a concise title for the pull-request.
- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)

# TL;DR
 - dynamic spec load failure should cause a system retry
 - Workqueue metrics name fixed
 - Better logging

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
dynamic spec load failure:
There are cases in which the Metastore (S3/GCS etc) may be unavailable for a small period of time - timeouts, network partition etc. When dynamic spec is loaded, if the spec does not exist, we immediately fail. This should only cause failure if the spec is incorrect - User error. For all system errors, we will retry

## Tracking Issue
https://github.com/flyteorg/flyte/issues/894


cmd/controller/cmd/root.go
```diff
@@ -69,8 +71,9 @@ var rootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	version.LogBuildInformation(appName)
+	logrus.Infof("Detected: %d CPU's\n", runtime.NumCPU())
```
EngHabu, line 74: why not stdlib/logger?

kumare3: I think its not initialized yet? The version code does not seem to use it?

kumare3: done!

EngHabu (approved):

PR #254, opened 2021-04-13
Cache dynamic workflow

katrogan:
# TL;DR
Cache dynamic workflow. Verified the cache is read during execution.

Also sends workflow spec to admin

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/899
https://github.com/flyteorg/flyte/issues/831
https://github.com/flyteorg/flyte/issues/928

## Follow-up issue
_NA_


pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -132,23 +133,32 @@ func (d dynamicNodeTaskNodeHandler) buildContextualDynamicWorkflow(ctx context.C
 	dynamicNodeStatus.SetOutputDir(nCtx.NodeStatus().GetOutputDir())
 	dynamicNodeStatus.SetParentTaskID(execID)
 
-	// cacheHitStopWatch := d.metrics.CacheHit.Start(ctx)
+	cacheHitStopWatch := d.metrics.CacheHit.Start(ctx)
```
kumare3, line 136: can you update a unit test for this?

katrogan: done

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -132,23 +133,32 @@ func (d dynamicNodeTaskNodeHandler) buildContextualDynamicWorkflow(ctx context.C
...
+			d.metrics.CacheError.Inc(ctx)
+		} else {
+			cacheHitStopWatch.Stop()
+			newParentInfo, err := node_common.CreateParentInfo(nCtx.ExecutionContext().GetParentInfo(), nCtx.NodeID(), nCtx.CurrentAttempt())
+			if err != nil {
+				return dynamicWorkflowContext{}, errors.Wrapf(utils.ErrorCodeSystem, err, "failed to generate uniqueID")
+			}
+			return dynamicWorkflowContext{
+				isDynamic:   true,
+				subWorkflow: compiledWf,
```
EngHabu, line 156: Shouldn't we set `subWorkflowClosure` here as well?

pkg/controller/nodes/dynamic/dynamic_workflow.go
EngHabu, line 136: Should we have a CacheMiss?

pkg/controller/nodes/dynamic/dynamic_workflow_test.go
```diff
@@ -329,4 +333,88 @@ func Test_dynamicNodeHandler_buildContextualDynamicWorkflow_withLaunchPlans(t *t
 		assert.Error(t, err)
 		assert.True(t, callsAdmin)
 	})
+	t.Run("dynamic wf cached", func(t *testing.T) {
```
kumare3, line 336: can we write a test for the failure case? Where the cache read fails, or the cache write fails

kumare3:
I am assuming you do not have the event sending in this PR right?

pkg/controller/nodes/dynamic/dynamic_workflow.go
katrogan: we don't cache that though, right? (if i'm understanding correctly)

katrogan:
@kumare3 we do send the event in this pr: https://github.com/flyteorg/flytepropeller/pull/254/files#diff-8cd2a8dbcbdb388d0af5b00672a4b005f27f3fda603d1dc9fd8b9ff4c8dd440bR135

pkg/controller/nodes/dynamic/dynamic_workflow.go
EngHabu: Do we send the dynamic workflow spec only the first time? or on every event (that will be too much)?
If it's only needed the first time, then I guess that's alright...

pkg/controller/nodes/dynamic/handler.go
```diff
@@ -130,7 +130,14 @@ func (d dynamicNodeTaskNodeHandler) handleDynamicSubNodes(ctx context.Context, n
 
 			return trns.WithInfo(handler.PhaseInfoFailureErr(ee.ExecutionError, trns.Info().GetInfo())), handler.DynamicNodeState{Phase: v1alpha1.DynamicNodePhaseFailing, Reason: ee.ExecutionError.String()}, nil
 		}
-		trns.WithInfo(trns.Info().WithInfo(&handler.ExecutionInfo{TaskNodeInfo: &handler.TaskNodeInfo{TaskNodeMetadata: &event.TaskNodeMetadata{CacheStatus: status.GetCacheStatus(), CatalogKey: status.GetMetadata()}}}))
+		taskNodeInfoMetadata := &event.TaskNodeMetadata{CacheStatus: status.GetCacheStatus(), CatalogKey: status.GetMetadata()}
+		if dCtx.subWorkflowClosure != nil && dCtx.subWorkflowClosure.Primary != nil && dCtx.subWorkflowClosure.Primary.Template != nil {
+			taskNodeInfoMetadata.DynamicWorkflow = &event.DynamicWorkflowNodeMetadata{
```
EngHabu, line 135: This is pretty static (it gets generated once and never changes for the duration of that node)... should we keep sending it to Admin on every event? isn't that too much?

kumare3: I would prefer we send it only once - does this need a state?

pkg/controller/nodes/dynamic/dynamic_workflow.go
katrogan: Yeah therefore we only send the spec the first time (when it's uncached) and don't need to bog down subsequent messages

pkg/controller/nodes/dynamic/handler.go
katrogan: yes we only send it once. for cached reads to the closure we don't send it

pkg/controller/nodes/dynamic/dynamic_workflow_test.go
katrogan: done

pkg/controller/nodes/dynamic/dynamic_workflow.go
katrogan: done

pkg/controller/nodes/dynamic/handler.go
EngHabu: I don't think we should rely on caching to not send it... can we make it based on the phase maybe? we only send it on queued or something like that... ? if there isn't a good existing node phase to control this, we can introduce one [here](https://github.com/flyteorg/flytepropeller/blob/master/pkg/apis/flyteworkflow/v1alpha1/node_status.go#L81-L92)...

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -116,6 +120,19 @@ func (d dynamicNodeTaskNodeHandler) buildDynamicWorkflowTemplate(ctx context.Con
 	}, nil
 }
 
+// Updates a dynamically generated subworkflow version to be deterministically generated from the computed closure.
+func setDeterministicVersion(ctx context.Context, workflowClosure *core.CompiledWorkflowClosure) error {
```
kumare3, line 124: this does not need to be done any more right? Why is this needed?

kumare3 (dismissed):

katrogan:
PTAL @EngHabu

pkg/controller/nodes/dynamic/dynamic_workflow.go
katrogan: reverted

EngHabu (approved):
Thank you for the verification runs!

PR #264, opened 2021-05-26
Change branch compilation to avoid adding downstream deps

EngHabu:

# TL;DR
- [X] Change branch compilation to only add subnodes during the transformer phase as opposed to compilation phase,
- [X] Avoid adding downstream links between branch nodes and their subnodes,
- [X] Marshal both downstream and upstream edges in propeller's WF CRD

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1052


pkg/controller/nodes/predicate.go
```diff
@@ -66,6 +68,8 @@ func CanExecute(ctx context.Context, dag executors.DAGStructure, nl executors.No
 				return PredicatePhaseUndefined, errors.Errorf(errors.BadSpecificationError, nodeID, "Upstream node [%v] of node [%v] not defined", upstreamNodeID, nodeID)
 			}
 
+			// Deprecated: This if block will be removed in a future version. It's harmless (will be no-op) for newly
```
kumare3, line 71: Good catch!
But question, is it really harmless? 
So we will have ParentNodeID set, but branch node will not be an upstream. If so yes, this is harmless

kumare3:
needs a merge!

kumare3 (approved):

kumare3 (approved):

PR #276, opened 2021-06-11
Compiler Enum Support and updated flyteidl #minor

kumare3:
# TL;DR
Support for ENUM types in flytecompiler and flytepropeller update to support it

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1rf0hMKbPebJ7jGYpsTDAbJ4ka58KIKD7EgWAEBIH06Q/edit#

## Tracking Issue
https://github.com/flyteorg/flyte/issues/590

## Follow-up issue
_NA_


pkg/compiler/validators/typing.go
```diff
@@ -33,6 +33,20 @@ func (t trivialChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
...
 		return true
 	}
+
+	// If upstream is an enum, it can be consumed as a string downstream
+	if upstreamType.GetEnumType() != nil {
+		if t.literalType.GetSimple() == flyte.SimpleType_STRING {
+			return true
+		}
+	}
+	// If t is an enum, it can be created from a string as Enums as just constrained String aliases
```
katrogan, line 43: is this a runtime error then if the string isn't a valid enum value?

katrogan (approved):

pkg/compiler/validators/typing.go
kumare3: yup

kumare3: but our entry points will not allow bad values

kumare3: @katrogan thats actually a good point. Can we restrict this in flyteadmin? Where we construct the launchplan?

katrogan: shouldn't the launch plan have to reference the enum type to be valid?

pkg/compiler/validators/typing.go
```diff
@@ -33,6 +33,20 @@ func (t trivialChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
 	if isVoid(upstreamType) {
 		return true
 	}
+
+	// If upstream is an enum, it can be consumed as a string downstream
+	if upstreamType.GetEnumType() != nil {
```
EngHabu, line 38: We should also add a check to the static binding to fail if the string isn't in the list...

PR #277, opened 2021-06-12
Compiler validates bound value for enum #patch

kumare3:

# TL;DR
The PR ensures that a bound value for Enums is valid at registration
/ compile time

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1rf0hMKbPebJ7jGYpsTDAbJ4ka58KIKD7EgWAEBIH06Q/edit#

## Tracking Issue
https://github.com/flyteorg/flyte/issues/590

## Follow-up issue
_NA_

### DIFF
--- a/pkg/compiler/validators/bindings.go
+++ b/pkg/compiler/validators/bindings.go
@@ -91,6 +91,21 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 			errs.Collect(errors.NewMismatchingTypesErr(nodeID, nodeParam, literalType.String(), expectedType.String()))
 		}
 
+		if expectedType.GetEnumType() != nil {
+			v := binding.GetScalar().GetPrimitive().GetStringValue()
+			// Let us assert that the bound value is a correct enum Value
+			found := false
+			for _, ev := range expectedType.GetEnumType().Values {
+				if ev == v {
+					found = true
+					break
+				}
+			}
+			if !found {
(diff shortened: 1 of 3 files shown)


pkg/compiler/validators/bindings.go
```diff
@@ -91,6 +91,21 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 			errs.Collect(errors.NewMismatchingTypesErr(nodeID, nodeParam, literalType.String(), expectedType.String()))
 		}
 
+		if expectedType.GetEnumType() != nil {
+			v := binding.GetScalar().GetPrimitive().GetStringValue()
```
katrogan, line 95: do you need to do any nil checks before dereferencing the string value?

katrogan: furthermore, what's the context of this call? does a binding always have an assigned value?

kumare3: yes, this is static binding. I was thinking you should not need a nil check, but I think you are right, it is possible that someone may try to add an int to an enum

kumare3: Actually I do not need the nil check. the AreTypesCastable will fail for any non string scalar when the source type is Enum

katrogan (approved):

EngHabu (approved):

PR #282, opened 2021-06-25
Use node resource overrides

katrogan:
Signed-off-by: Katrina Rogan <katroganGH@gmail.com>

# TL;DR
Implements node resource overrides.

Side note, we should find a good way to explore exposing these overrides to the user since the task spec in the console will only show the task resource requirements, rather than the resolved value.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1170

## Follow-up issue
_NA_


EngHabu (approved):
sweet

PR #288, opened 2021-07-12
Fix case in which downstream nodes may incorrectly fail because of parentNodeID #minor

kumare3:
# TL;DR
Downstream nodes of a node, have the parentNodeID always set. For the branch node, the parent node is the branch node of all the branches. When a sub-node of branch is to be executed, it should wait for the branch to be success, but the branch node (as it is the container node) will not be marked as success.
A branch sub-node may also be downstream from a regular node (e.g. start-node). Thus logically it can be executed after this node. But, it should wait for the branch condition to be evaluated and for the node to be ready.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1232

### DIFF
--- a/pkg/controller/nodes/predicate.go
+++ b/pkg/controller/nodes/predicate.go
@@ -56,15 +56,15 @@ func CanExecute(ctx context.Context, dag executors.DAGStructure, nl executors.No
 
 	skipped := false
 	for _, upstreamNodeID := range upstreamNodes {
+		upstreamNode, ok := nl.GetNode(upstreamNodeID)
+		if !ok {
+			return PredicatePhaseUndefined, errors.Errorf(errors.BadSpecificationError, nodeID, "Upstream node [%v] of node [%v] not defined", upstreamNodeID, nodeID)
+		}
+
 		upstreamNodeStatus := nl.GetNodeExecutionStatus(ctx, upstreamNodeID)
 
 		if upstreamNodeStatus.IsDirty() {
 			return PredicatePhaseNotReady, nil
-		}
-
-		upstreamNode, ok := nl.GetNode(upstreamNodeID)
-		if !ok {
-			return PredicatePhaseUndefined, errors.Errorf(errors.BadSpecificationError, nodeID, "Upstream node [%v] of node [%v] not defined", upstreamNodeID, nodeID)
 		}
 
 		if upstreamNode.GetBranchNode() != nil && upstreamNodeStatus.GetBranchStatus() != nil {
--- a/pkg/controller/nodes/predicate.go
+++ b/pkg/controller/nodes/predicate.go
@@ -38,6 +38,8 @@ func (p PredicatePhase) String() string {
 	return "undefined"
 }
 
+// CanExecute method informs the callee if the given node can begin execution. This is dependent on
+// primarily that all nodes upstream to the given node are successful and the results are available.
 func CanExecute(ctx context.Context, dag executors.DAGStructure, nl executors.NodeLookup, node v1alpha1.BaseNode) (
 	PredicatePhase, error) {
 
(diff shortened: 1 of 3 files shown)


pkg/controller/nodes/predicate.go
```diff
@@ -62,21 +62,21 @@ func CanExecute(ctx context.Context, dag executors.DAGStructure, nl executors.No
...
-				logger.Debugf(ctx, "Branch sub node is expected to have parent branch node in succeeded state")
-				return PredicatePhaseUndefined, errors.Errorf(errors.IllegalStateError, nodeID, "Upstream node [%v] is set as parent, but is not a branch node of [%v] or in illegal state.", upstreamNodeID, nodeID)
+		if upstreamNode.GetBranchNode() != nil && upstreamNodeStatus.GetBranchStatus() != nil {
+			if upstreamNodeStatus.GetBranchStatus().GetPhase() != v1alpha1.BranchNodeSuccess {
+				return PredicatePhaseNotReady, nil
+			}
+			// Branch node is success, so we are free to go ahead and execute the child nodes of the branch node, but
+			// not any of the dependent nodes
+			nodeStatus := nl.GetNodeExecutionStatus(ctx, nodeID)
+			if nodeStatus.GetParentNodeID() != nil && *nodeStatus.GetParentNodeID() == upstreamNodeID {
```
EngHabu, line 77: Maybe add:
```suggestion
			// ParentNodeID is set for sub-workflows as well as branch nodes. Between these two cases, only branch nodes will have upstreamNodeId set. If that's the case, we do not need to further check the status of the upstream node.
			if nodeStatus.GetParentNodeID() != nil && *nodeStatus.GetParentNodeID() == upstreamNodeID {
```

pkg/controller/nodes/predicate.go
```diff
@@ -62,21 +62,21 @@ func CanExecute(ctx context.Context, dag executors.DAGStructure, nl executors.No
...
+			return PredicatePhaseUndefined, errors.Errorf(errors.BadSpecificationError, nodeID, "Upstream node [%v] of node [%v] not defined", upstreamNodeID, nodeID)
+		}
 
-			// Deprecated: This if block will be removed in a future version. It's harmless (will be no-op) for newly
-			// compiled Workflows as sub-branch-nodes won't have an execution or code dependency on branch nodes.
-			// This only happens if current node is the child node of a branch node
-			if upstreamNode.GetBranchNode() == nil || upstreamNodeStatus.GetBranchStatus().GetPhase() != v1alpha1.BranchNodeSuccess {
-				logger.Debugf(ctx, "Branch sub node is expected to have parent branch node in succeeded state")
-				return PredicatePhaseUndefined, errors.Errorf(errors.IllegalStateError, nodeID, "Upstream node [%v] is set as parent, but is not a branch node of [%v] or in illegal state.", upstreamNodeID, nodeID)
+		if upstreamNode.GetBranchNode() != nil && upstreamNodeStatus.GetBranchStatus() != nil {
```
EngHabu, line 70: I think the condition is still puzzling... I understand what it's trying to do but I think we should rephrase it like below. The issue I take is that we check if upstream is a branch node... which technically we don't care about UNLESS it's also the parent node... I understand the condition will work either way but it's just odd to special case branch node in that case... does that makes sense?

```
If I've a parentId and it's the same as the upstreamNode:
   If it's a branch node, and status is not BranchNodeSuccess, return not ready
   If it's a branch node, and status is BranchNodeSuccess, continue
   If it's NOT a branch node, error? should never happen? 
Else: check upstream has succeeded... etc. etc.
```

EngHabu (commented):
why `#minor`?

EngHabu (commented):
Let's document the code in the predicate?

EngHabu (approved):

PR #290, opened 2021-07-13
Implement workflow execution recovery

katrogan:
# TL;DR
Implement workflow execution recovery. For ordinary task nodes this modifies pre-execute to attempt to recover node executions when the overall workflow execution is run in recovery mode, helping elide needless re-computation for previously succeeded node executions.

For workflow nodes (that is, those that call out to create an execution from a launch plan) this sets the triggered executions to recover from the original node execution which created an execution using the same launch plan.

For dynamic nodes (that is those that contain a subworkflow) the behavior remains the same as task nodes. If the original dynamic node succeeded, great, we'll recover it. Otherwise the dynamic workflow runs again.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1151

## Follow-up issue
_NA_


pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -92,6 +94,10 @@ func (p NodePhase) String() string {
 		return "RetryableFailure"
 	case NodePhaseDynamicRunning:
 		return "DynamicRunning"
+	case NodePhaseRecovering:
```
kumare3, line 97: you have to do op_code_generate

pkg/controller/nodes/subworkflow/launchplan/admin_test.go
```diff
@@ -193,6 +195,42 @@ func TestAdminLaunchPlanExecutor_Launch(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
+	t.Run("happy recover", func(t *testing.T) {
```
kumare3, line 198: what if recover fails?

pkg/controller/nodes/executor.go
```diff
@@ -146,6 +152,119 @@ func (c *nodeExecutor) IdempotentRecordEvent(ctx context.Context, nodeEvent *eve
...
+	if recoveredData.FullOutputs != nil {
+		outputs = recoveredData.FullOutputs
+	} else if len(recovered.Closure.GetOutputUri()) > 0 {
+		if err := c.store.ReadProtobuf(ctx, storage.DataReference(recovered.Closure.GetOutputUri()), outputs); err != nil {
+			return handler.PhaseInfoUndefined, errors.Wrapf(errors.InputsNotFoundError, nCtx.NodeID(), err, "failed to read output data [%v].", recovered.Closure.GetOutputUri())
+		}
+	} else {
+		logger.Debugf(ctx, "No outputs found for recovered node [%+v]", nCtx.NodeExecutionMetadata().GetNodeExecutionID())
+	}
+	outputFile := v1alpha1.GetOutputsFile(nCtx.NodeStatus().GetOutputDir())
```
kumare3, line 234: ohh no are we still using this?

pkg/controller/nodes/executor.go
```diff
@@ -429,14 +554,19 @@ func (c *nodeExecutor) handleQueuedOrRunningNode(ctx context.Context, nCtx *node
 		np = v1alpha1.NodePhaseSucceeded
 		finalStatus = executors.NodeStatusSuccess
 	}
+	if np == v1alpha1.NodePhaseRecovering && !h.FinalizeRequired() {
```
kumare3, line 557: if it is recovering we do not even care about the  handler or finalization right?

kumare3: do we really need `Recovering`?

kumare3:
Its amazing how clean this change is. It really fits in well into the premise?

pkg/controller/controller.go
```diff
@@ -419,9 +422,20 @@ func New(ctx context.Context, cfg *config.Config, kubeclientset kubernetes.Inter
 
 	controller.levelMonitor = NewResourceLevelMonitor(scope.NewSubScope("collector"), flyteworkflowInformer.Lister())
 
+	// The admin client might not be initialized if EnableAdminLauncher is set to False.
+	if adminClient == nil {
+		var err error
+		adminClient, err = getAdminClient(ctx)
```
EngHabu, line 428: If we now always want an admin client, let's remove the condition and just always initialize it and use it up there if needed and down here all the time...

pkg/controller/nodes/subworkflow/launchplan.go
```diff
@@ -17,7 +21,8 @@ import (
 )
 
 type launchPlanHandler struct {
-	launchPlan launchplan.Executor
+	launchPlan     launchplan.Executor
+	recoveryClient recovery.Client
```
EngHabu, line 25: I'm a bit confused about why do we need the recovery client at this layer... won't that be handled at the Node executor layer?

kumare3: Isn't the reason that we want to call the recover execution endpoint instead in case the child workflow node has failed, so that we can recover a partially failed child workflow?

EngHabu: That's what I thought.. but looking at the interface and implementation of recovery client... it retrieves Node executions from admin... I think... I obviously could be missing something since I'm half looking at this and half prepping my bags :-D

pkg/controller/nodes/executor.go
katrogan: @kumare3 what should i be using instead?

pkg/controller/nodes/subworkflow/launchplan.go
katrogan: yes, we retrieve the node execution - which has target metadata of type workflow node, which we can use to fetch the originally-launched child execution

pkg/controller/nodes/executor.go
katrogan: updated

pkg/controller/nodes/subworkflow/launchplan/admin_test.go
katrogan: updated to call create execution when applicable

pkg/controller/controller.go
katrogan: done

katrogan:
PTAL @kumare3

pkg/controller/nodes/executor.go
kumare3: i think we use this now - https://github.com/flyteorg/flyteplugins/blob/60b94c688ef2b98aa53a9224b529ac672af04540/go/tasks/pluginmachinery/ioutils/paths.go

katrogan: @kumare3 this doesn't include output file paths though. should i update it?

kumare3 (approved):

PR #292, opened 2021-07-13
Construct subnode DataDir to be under parent's node OutputDir to keep behavior consistent across

EngHabu:

# TL;DR
When constructing a node status, the `DataDir` was incorrectly set to be a child of the parent's `DataDir` when it should have been a child of the parent's `OutputDir`.

This, incidentally, only broke in the case of Dynamic Workflow caching because in all other cases we explicitly set the data directory (essentially overriding the bad value retrieved from `GetNodeExecutionStatus`).

Further cleanup/refactor is needed to avoid having multiple branches of the code that set up data dir differently.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1238


katrogan (approved):

milton0825 (approved):

EngHabu:
😢

EngHabu:
⏪

PR #293, opened 2021-07-15
Revert "Construct subnode DataDir to be under parent's node OutputDir to keep behavior consistent across"

EngHabu:
Reverts flyteorg/flytepropeller#292

### DIFF
--- a/pkg/apis/flyteworkflow/v1alpha1/node_status.go
+++ b/pkg/apis/flyteworkflow/v1alpha1/node_status.go
@@ -523,7 +523,7 @@ func (in *NodeStatus) GetNodeExecutionStatus(ctx context.Context, id NodeID) Exe
 		n.SetParentTaskID(in.GetParentTaskID())
 		n.DataReferenceConstructor = in.DataReferenceConstructor
 		if len(n.GetDataDir()) == 0 {
-			dataDir, err := in.DataReferenceConstructor.ConstructReference(ctx, in.GetOutputDir(), id)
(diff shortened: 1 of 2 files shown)


pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -523,7 +523,7 @@ func (in *NodeStatus) GetNodeExecutionStatus(ctx context.Context, id NodeID) Exe
 		n.SetParentTaskID(in.GetParentTaskID())
 		n.DataReferenceConstructor = in.DataReferenceConstructor
 		if len(n.GetDataDir()) == 0 {
-			dataDir, err := in.DataReferenceConstructor.ConstructReference(ctx, in.GetOutputDir(), id)
+			dataDir, err := in.DataReferenceConstructor.ConstructReference(ctx, in.GetDataDir(), id)
```
milton0825, line 526: Shouldn't it be
```suggestion
			dataDir, err := in.DataReferenceConstructor.ConstructReference(ctx, in.GetOutputDir(), id)
```

milton0825: I tried https://github.com/flyteorg/flytepropeller/pull/292/commits/8310d4fb822c5f97b90216f3209f55ccf5d788e3 in lyft staging and it fixed the issue

EngHabu: Yeah it's just this is "magic" code that I don't want to have around... super hard to maintain... so let's revert this and understand the issue and fix it properly...

kumare3 (approved):

PR #294, opened 2021-07-15
Simplify GetNodeExecutionStatus

EngHabu:

# TL;DR
For ev 0, parentnodeid is not important except for branch nodes. In case of dynamic workflow caching this was not set correctly. This PR changes the behavior and also streamlines the code block

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
For event version-0 the dynamic array task did not generate the right node ids and relationship. This change fixes it

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1236


pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -517,62 +518,53 @@ func (in NodeStatus) GetTaskNodeStatus() ExecutableTaskNodeStatus {
...
-			if err != nil {
-				logger.Errorf(ctx, "Failed to construct output dir for node [%v]", id)
-				return n
-			}
+		n.SetDataDir(dataDir)
+	}
 
-			n.SetOutputDir(outputDir)
+	if len(n.GetOutputDir()) == 0 {
+		outputDir, err := in.DataReferenceConstructor.ConstructReference(ctx, n.GetDataDir(), strconv.FormatUint(uint64(n.Attempts), 10))
```
katrogan, line 533: why is this using n.GetDataDir to set the outputDir and above in line 524 the OutputDir to set the dataDir. This seems backwards

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -171,27 +201,53 @@ func (d dynamicNodeTaskNodeHandler) buildContextualDynamicWorkflow(ctx context.C
...
+		return workflowContext, err
+	}
+
+	if err := f.Cache(ctx, dynamicWf, closure); err != nil {
+		logger.Errorf(ctx, "Failed to cache Dynamic workflow [%s]", err.Error())
+	}
+
+	// The current node would end up becoming the parent for the dynamic task nodes.
+	// This is done to track the lineage. For level zero, the CreateParentInfo will return nil
+	newParentInfo, err := node_common.CreateParentInfo(nCtx.ExecutionContext().GetParentInfo(), nCtx.NodeID(), nCtx.CurrentAttempt())
```
katrogan, line 215: do we not need to create the new parent info in the cache read case too?

pkg/controller/nodes/dynamic/dynamic_workflow.go
```diff
@@ -171,27 +201,53 @@ func (d dynamicNodeTaskNodeHandler) buildContextualDynamicWorkflow(ctx context.C
...
+	closure, dynamicWf, workflowContext, err := d.buildDynamicWorkflow(ctx, nCtx, djSpec, dynamicNodeStatus)
+	if err != nil {
+		return workflowContext, err
+	}
+
+	if err := f.Cache(ctx, dynamicWf, closure); err != nil {
+		logger.Errorf(ctx, "Failed to cache Dynamic workflow [%s]", err.Error())
+	}
+
+	// The current node would end up becoming the parent for the dynamic task nodes.
```
katrogan, line 213: do you need to call setEphemeralNodeExecutionStatusAttributes here too?

pkg/apis/flyteworkflow/v1alpha1/node_status.go
kumare3: since the datadir is already set, output dir is just a derivative from that

pkg/controller/nodes/dynamic/dynamic_workflow.go
kumare3: in the case of the cache, sadly, GetNodeExecutionStatus has a side-effect that sets those attributes

kumare3 (approved):

PR #310, opened 2021-08-24
Resolve resource requirements in propeller at execution time

katrogan:
Signed-off-by: Katrina Rogan <katroganGH@gmail.com>

# TL;DR
Numerous issues keep cropping up because task resource resolution is convoluted and handled in multiple repos at multiple points in the registration -> execution lifecycle.

FlyteAdmin currently injects resource overrides for container tasks, but as we move to more complex targets it makes more sense for propeller to reconcile node overrides with platform-specified requests/limits and then have the plugin handler make sense of these at execution time.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1361

## Follow-up issue
_NA_


pkg/apis/flyteworkflow/v1alpha1/execution_config.go
```diff
@@ -24,9 +24,23 @@ type ExecutionConfig struct {
...
+	// Defines the resource requests and limits specified for this task that ought to be applied at execution time.
+	TaskResources TaskResources
 }
 
 type TaskPluginOverride struct {
 	PluginIDs             []string
 	MissingPluginBehavior admin.PluginOverride_MissingPluginBehavior
 }
+
+type TaskResourceSpec struct {
```
EngHabu, line 36: docs plz

pkg/apis/flyteworkflow/v1alpha1/execution_config.go
```diff
@@ -24,9 +24,23 @@ type ExecutionConfig struct {
...
+	CPU              string
+	Memory           string
+	EphemeralStorage string
+	Storage          string
```
EngHabu, line 40: Should we use `resource.Quantity`?

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -178,7 +252,7 @@ func (t *Handler) newTaskExecutionContext(ctx context.Context, nCtx handler.Node
 		tm: taskExecutionMetadata{
 			NodeExecutionMetadata: nCtx.NodeExecutionMetadata(),
 			taskExecID:            taskExecutionID{execName: uniqueID, id: id},
-			o:                     nCtx.Node(),
+			o:                     newTaskOverrides(nCtx.Node(), determineResourceRequirements(nCtx, nCtx.ExecutionContext().GetExecutionConfig())),
```
EngHabu, line 255: ?
```suggestion
			o:                     newTaskOverrides(nCtx.Node(), determineResourceRequirements(nCtx, nCtx.ExecutionContext().GetExecutionConfig().TaskResources)),
```

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -124,6 +143,61 @@ func (t taskExecutionContext) SecretManager() pluginCore.SecretManager {
...
 }
 
+// Validates and assigns a single resource by examining the default requests and max limit with the static resource value
+// defined by this task and node execution context.
+func assignResource(resourceName v1.ResourceName, execConfigRequest, execConfigLimit string, requests, limits v1.ResourceList) {
+	if len(execConfigLimit) == 0 || len(execConfigRequest) == 0 {
+		return
+	}
+	request, ok := requests[resourceName]
+	maxLimit := resource.MustParse(execConfigLimit)
```
EngHabu, line 153: I would rather avoid panics if we can help it...

```suggestion
	maxLimit, err := resource.Parse(execConfigLimit)
	if err != nil {
	  return err
	}
	
	request, ok := requests[resourceName]
```

EngHabu: In fact, if we change the CRD type to resource.Quantity, we don't need to do parsing here at all...

katrogan:
PTAL @EngHabu

EngHabu (approved):
Thank you!

PR #317, opened 2021-09-07
Start-Node Output file copied to output result

pmahindrakar-oss:
Signed-off-by: Prafulla Mahindrakar <prafulla.mahindrakar@gmail.com>

# TL;DR
The node data returned for start-node is currently incorrect from admin and this is due to the fact that propeller is sending non existent inputUri for  start-node

Flytekit gets around this issue by skipping the call to admin altogether and similar fix was done in flytectl 
https://github.com/flyteorg/flytekit/blob/master/flytekit/clis/flyte_cli/main.py#L1540

This is an attempt to fix the issue in propeller.
For start-node the Input and Output data is the same and hence using this fact we send the node event from propeller.

After the fix 
```
[
	{
		"node_exec": {
			"id": {
				"nodeId": "start-node",
				"executionId": {
					"project": "flytesnacks",
					"domain": "development",
					"name": "fdbbaa49c298c4bb48ad"
				}
			},
			"closure": {
				"outputUri": "s3://my-s3-bucket/metadata/propeller/sandbox/flytesnacks-development-fdbbaa49c298c4bb48ad/start-node/data/0/outputs.pb",
				"phase": "SUCCEEDED",
				"createdAt": "2021-09-14T11:51:04.355307Z",
				"updatedAt": "2021-09-14T11:51:04.355307Z"
			},
			"metadata": {
				"specNodeId": "start-node"
			}
		},
		"outputs": {
			"am": true,
			"day_of_week": "Sunday",
			"number": 5
		}
	}
]
```

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1221

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/nodes/transformers.go
+++ b/pkg/controller/nodes/transformers.go
@@ -95,14 +95,13 @@ func ToNodeExecutionEvent(nodeExecID *core.NodeExecutionIdentifier,
 	}
 
 	var nev *event.NodeExecutionEvent
-	// Start node is special case where the Inputs and Outputs are the same and hence here we copy the Output info
-	// in both the InputUri and OutputUri
+	// Start node is special case where the Inputs and Outputs are the same and hence here we copy the Output file
+	// into the OutputResult and in admin we copy it over into input aswell.
 	if nodeExecID.NodeId == v1alpha1.StartNodeID {
 		outputsFile := v1alpha1.GetOutputsFile(status.GetOutputDir())
 		nev = &event.NodeExecutionEvent{
 			Id:       nodeExecID,
 			Phase:    phase,
-			InputUri: outputsFile.String(),
 			OutputResult: ToNodeExecOutput(&handler.OutputInfo{
 				OutputURI: outputsFile,
 			}),
--- a/pkg/controller/nodes/transformers.go
+++ b/pkg/controller/nodes/transformers.go
@@ -100,8 +100,8 @@ func ToNodeExecutionEvent(nodeExecID *core.NodeExecutionIdentifier,
 	if nodeExecID.NodeId == v1alpha1.StartNodeID {
 		outputsFile := v1alpha1.GetOutputsFile(status.GetOutputDir())
 		nev = &event.NodeExecutionEvent{
-			Id:       nodeExecID,
-			Phase:    phase,
+			Id:    nodeExecID,
+			Phase: phase,
 			OutputResult: ToNodeExecOutput(&handler.OutputInfo{
 				OutputURI: outputsFile,
 			}),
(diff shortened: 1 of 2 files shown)


pkg/controller/nodes/transformers.go
```diff
@@ -94,11 +94,24 @@ func ToNodeExecutionEvent(nodeExecID *core.NodeExecutionIdentifier,
...
-		Id:         nodeExecID,
-		Phase:      phase,
-		InputUri:   inputPath,
-		OccurredAt: occurredTime,
+	var nev *event.NodeExecutionEvent
+	if nodeExecID.NodeId == v1alpha1.StartNodeID {
+		nev = &event.NodeExecutionEvent{
+			Id:       nodeExecID,
+			Phase:    phase,
+			InputUri: v1alpha1.GetOutputsFile(status.GetOutputDir()).String(),
```
kumare3, line 102: move the `v1alpha1.GetOutputsFile(status.GetOutputDir()).String()` to a common var in the if block? This will make it clear that outputs are being copied to both inputs and outputs. Also, can you write a comment about this?

@katrogan, if we send the entire data within the payload, would the input and output uris be unfurled each time. If so should we move this peculiar logic to Admin?

kumare3:
LGTM, after the one comment and info from @katrogan

pkg/controller/nodes/transformers.go
pmahindrakar-oss: So right the OutputUri is not even sent for the start-node so bare minimum we would need a change to send that from propeller. And also that would be similar custom logic which checks for the start-node and populates the OutputUri or InputUri in propeller and then admin can just copy over the value with another custom logic there for the start-node to other field.

katrogan: @kumare3 if we send the URIs in the payload admin will not unfurl them unless someone calls for GetNodeExecutionData - which itself perhaps ought to be (lazily) cached somehow

kumare3: @katrogan i was referring to https://github.com/flyteorg/flytepropeller/blob/61893990bb8d0a46133ce0cc81b94f9c1e647764/pkg/controller/events/node_execution.go#L53

katrogan: @kumare3 we can't always unfurl in admin, right? because the whole purpose of setting the policy at propeller side is to support a distributed propeller with protected s3 access

EngHabu (approved):

kumare3 (approved):

PR #320, opened 2021-09-09
gracefully failing workflow on ExecutionNotFound error

hamersaw:
Signed-off-by: Daniel Rammer <daniel@union.ai>

# TL;DR
Gracefully fail on ExecutionNotFound by transitioning the workflow the the 'FAILING' state. Subsequently, propeller will abort workflow tasks and transition the phase to 'FAILED'.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This facilitates graceful failures when experiencing an ExecutionNotFound error during workflow transitions. The error occurs when flyteadmin either doesn't have the provided workflow or the workflow is deleted. Either case is an invalid state and should result in the failure of the specified workflow.
We handle the race condition where flyteadmin, during the two phase execution, is between initializing the CRD and writing to the DB by checking if the workflow is in the 'READY' phase.
By transitioning the workflow to the 'FAILING' phase propeller will attempt to abort all active tasks on subsequent processing before transitioning to the 'FAILED' phase and concluding processing. This approach is similar to the failing status transition when a flyte workflow is too large to update in the wfStore. 

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1048

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/handler.go
+++ b/pkg/controller/handler.go
@@ -230,9 +230,9 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 			}
 		}
 
-		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not a valid state
-		// unless we are experiencing a race condition where the workflow has not yet been inserted
-		// into the db (ie. workflow phase is WorkflowPhaseReady).
+		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not
+		// a valid state unless we are experiencing a race condition where the workflow has not yet
+		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
 		if err != nil {
 			e, ok := err.(*errors.WorkflowErrorWithCause)
 			if ok && eventsErr.IsNotFound(e.Cause()) && w.GetExecutionStatus().GetPhase() != v1alpha1.WorkflowPhaseReady {
--- a/pkg/controller/handler.go
+++ b/pkg/controller/handler.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"runtime/debug"
@@ -15,7 +16,6 @@ import (
 
 	"github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 	"github.com/flyteorg/flytepropeller/pkg/controller/config"
-	"github.com/flyteorg/flytepropeller/pkg/controller/workflow/errors"
 	"github.com/flyteorg/flytepropeller/pkg/controller/workflowstore"
 
 	"github.com/flyteorg/flytestdlib/logger"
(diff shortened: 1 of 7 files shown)


pkg/controller/handler.go
```diff
@@ -227,6 +229,31 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
+		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
+		if err != nil {
+			e, ok := err.(*errors.WorkflowErrorWithCause)
+			if ok && eventsErr.IsNotFound(e.Cause()) && w.GetExecutionStatus().GetPhase() != v1alpha1.WorkflowPhaseReady {
+				t.Stop()
+				logger.Errorf(ctx, "Failed to process workflow, failing: %s", e.Cause())
+
+				// We set the workflow status to failing to abort any active tasks in the next round.
+				mutableW := w.DeepCopy()
+				mutableW.Status.UpdatePhase(v1alpha1.WorkflowPhaseFailing, "Workflow execution is missing in flyteadmin, aborting", &core.ExecutionError{
```
EngHabu, line 244: I forgot how we handle WorkflowPhaseFailing, do we try to send any events while handling that phase? if we do, those will also fail... and we will get stuck in that phase...

pkg/controller/handler.go
```diff
@@ -227,6 +229,31 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 				ResetFinalizers(mutatedWf)
 			}
 		}
+
+		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not
+		// a valid state unless we are experiencing a race condition where the workflow has not yet
+		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
+		if err != nil {
+			e, ok := err.(*errors.WorkflowErrorWithCause)
```
EngHabu, line 237: I would run the error through `stdErrors.Unwrap` first... in case it was wrapped somewhere on the stack...
```
import (
...
   stdErrors "errors"
)
```

EngHabu (commented):
Awesome! just a couple of clarifications

pkg/controller/handler.go
hamersaw: This is a very thrifty API and a very useful suggestion. It looks like errors.Unwrap function only unwraps a single layer though (and requires the Unwrap function defined on the internal error). Alternatively, I changed the check to use errors.Is directly on the ExecutionNotFound error which checks if it is anywhere in the stack. However, this does require a slight modification to the flyteidl repo to include EventError.Is and EventError.Unwrap functions. I will submit a pull request shortly.

pkg/controller/handler.go
hamersaw: You are correct, the WorkflowPhaseFailing does attempt to transition phase and send an event to admin. I added a check (and comment) in pkg/controller/workflow/executor.go to ignore the ExecutionNotFound error for the Failing to Failed transition. I think this makes sense because if flyteadmin is missing the execution we don't need to add that error to the list of reasons the workflow is failing, we can just gracefully fail. The error is still logged in multiple locations.

pkg/controller/handler.go
```diff
@@ -227,6 +228,28 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
...
+		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not
+		// a valid state unless we are experiencing a race condition where the workflow has not yet
+		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
+		if err != nil && eventsErr.IsNotFound(err) && w.GetExecutionStatus().GetPhase() != v1alpha1.WorkflowPhaseReady {
+			t.Stop()
+			logger.Errorf(ctx, "Failed to process workflow, failing: %s", err)
+
+			// We set the workflow status to failing to abort any active tasks in the next round.
+			mutableW := w.DeepCopy()
+			mutableW.Status.UpdatePhase(v1alpha1.WorkflowPhaseFailing, "Workflow execution is missing in flyteadmin, aborting", &core.ExecutionError{
```
kumare3, line 241: why failing? I think we should move to failed right, and also remove the finalizer etc?

hamersaw: I was setting to failing to leverage the existing graceful shutdown mechanisms during the transition from failing to failed. For example, propagating Abort throughout the workflow to execute KillTask on any active k8s resources. Is this not as critical as I imagined? It's certainly easier to set to failed and be done with it.

kumare3: https://docs.flyte.org/en/latest/concepts/state_machine.html#divedeep-state-machine

pkg/controller/nodes/errors/errors.go
```diff
@@ -23,11 +23,23 @@ func (n *NodeError) Error() string {
 	return fmt.Sprintf("failed at Node[%s]. %v: %v", n.Node, n.ErrCode, n.Message)
 }
 
+func (n *NodeError) Is(target error) bool {
+	t, ok := target.(*NodeError)
+	if !ok {
+		return false
+	}
+	return n.ErrCode == t.ErrCode && (n.Message == t.Message || t.Message == "") && (n.Node == t.Node || t.Node == "")
```
EngHabu, line 31: Same change needed as in the idl change?

pkg/controller/workflow/executor.go
```diff
@@ -405,7 +405,10 @@ func (c *workflowExecutor) HandleFlyteWorkflow(ctx context.Context, w *v1alpha1.
 			return err
 		}
 		if err := c.TransitionToPhase(ctx, w.ExecutionID.WorkflowExecutionIdentifier, wStatus, newStatus); err != nil {
-			return err
+			// Ignore ExecutionNotFound error to allow graceful failure
+			if err != nil && !eventsErr.IsNotFound(err) {
+				return err
+			}
```
EngHabu, line 411: There is one more failure scenario we need to handle here. There is something called FailureNode (which is sort of the finally{} block in a try/catch block that's only called on failure (Mainly to cleanup up resources... etc.).

If there is a FailureNode defined, it'll transition HandlingFailureNode which will then try to start the node (transition that to Queued which will fail) then it'll get stuck at that phase...

I think if you add this block to the next `switch case v1alpha1.WorkflowHandlingFailureNode` [below ](https://github.com/flyteorg/flytepropeller/blob/6a14e7fbffe89786fb1d8cde22715f93c2f3aff5/pkg/controller/workflow/executor.go#L418) that would be enough.

EngHabu (commented):
Excellent! just one comment

hamersaw:
@EngHabu tests will fail because the flyteidl version is still set to 0.19.22 and doesn't contain our recent fix to the eventErrors.IsNotFound function. Is it OK to increase the version in this PR? Or do we handle that with a separate process?

EngHabu:
> @EngHabu tests will fail because the flyteidl version is still set to 0.19.22 and doesn't contain our recent fix to the eventErrors.IsNotFound function. Is it OK to increase the version in this PR? Or do we handle that with a separate process?

Yes, feel free to go ahead and update flyteidl

kumare3:
@hamersaw you can test this on sandbox, by simply deleting the database and restarting it, or just deleting some executions in the table

hamersaw:
> @hamersaw you can test this on sandbox, by simply deleting the database and restarting it, or just deleting some executions in the table

sure, i've been testing by deploying flyte on a k3d cluster, scaling the flytepropeller deployment to 0, and starting a local instance with the correct connection configurations, then modifying the db as you suggested. i'm convinced everything is working.

kumare3 (approved):

EngHabu (approved):
Awesome!

PR #322, opened 2021-09-10
Separate out resource overrides from static resource definitions

katrogan:
# TL;DR
The introduction of https://github.com/flyteorg/flytepropeller/pull/310 means that platform resource defaults are now applied as task resource defaults in the absence of task decorator resource specifications. For container tasks, this is fine. For pod tasks this results in the pod spec primary container definition resource spec being overwritten by the platform defaults. 

Tested on flytesandbox

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1450

## Follow-up issue
_NA_


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
 	github.com/fatih/color v1.10.0
 	github.com/flyteorg/flyteidl v0.21.0
-	github.com/flyteorg/flyteplugins v0.6.0
+	github.com/flyteorg/flyteplugins v0.6.2-0.20210910212028-145d6e0a0d83
```
katrogan, line 10: will update before merging

pkg/apis/flyteworkflow/v1alpha1/nodes.go
```diff
@@ -93,17 +93,18 @@ func (in *NodeMetadata) DeepCopyInto(out *NodeMetadata) {
...
-	TaskRef       *TaskID                       `json:"task,omitempty"`
-	WorkflowNode  *WorkflowNodeSpec             `json:"workflow,omitempty"`
-	InputBindings []*Binding                    `json:"inputBindings,omitempty"`
-	Config        *typesv1.ConfigMap            `json:"config,omitempty"`
-	RetryStrategy *RetryStrategy                `json:"retry,omitempty"`
-	OutputAliases []Alias                       `json:"outputAlias,omitempty"`
+	ID                NodeID                        `json:"id"`
+	Name              string                        `json:"name,omitempty"`
+	Resources         *typesv1.ResourceRequirements `json:"resources,omitempty"`
+	ResourceOverrides *typesv1.ResourceRequirements `json:"resourceOverrides,omitempty"`
```
kumare3, line 99: this is getting complicated. Earlier we would compute the overrides in the compiler and send it here. Now we have both?

katrogan: sure, as an alternative we could compute resources and also keep track of platform defaults/limits and apply as needed in plugins

PR #324, opened 2021-09-14
Add support for defining status as a subresource of flyteworkflow CRD

hamersaw:
# TL;DR
Adding support for setting status as a subresource on the flyteworkflow CRD. This should reduce traffic to the kubernetes API server by reducing update message sizes.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
Setting status as a subresource on the flyteworkflow CRD introduces a separate kubernetes rest API which partitions updates of the CRD spec and status subfields. This functionality seems straight-forward, but introduces complexities when trying to provide backward compatible support for existing deployments where the CRD doesn't have the status defined as a subresource. Specifically, the cases are described below for both the existing and non-existing CRD status subresource and the effect of the Update and UpdateStatus API calls.

case 1: subresource exists on CRD
    1.1 client.Update(...) -> succeeds, but status does not change - update doesn't care about changing status subresource
    1.2 client.UpdateStatus(...) -> succeeeds
case 2: no subresouce on CRD
    2.1 -> client.Update(...) -> succeeds (status does change)
    2.2 -> client.UpdateStatus(...) -> throws k8s error NotFound  - impossible to distinguish from 1.2 where the resource does not actually exist

Therefore, we determine if the status subresource API is available and use it to inform the correct commands to update flyteworkflow CRD status and spec subfields. This means the implementation works for the existing flyteworkflow CRD definition (ie. without status subresource) and offers advantages once the subresource is added.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/126

## Follow-up issue
_NA_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/go-redis/redis v6.15.7+incompatible
 	github.com/go-test/deep v1.0.7
 	github.com/golang/protobuf v1.4.3
+	github.com/google/go-cmp v0.5.4
 	github.com/google/uuid v1.2.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.1.0
 	github.com/magiconair/properties v1.8.4
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -315,7 +315,7 @@ func getAdminClient(ctx context.Context) (client service.AdminServiceClient, err
 	return clients.AdminClient(), nil
 }
 
-// Check to see if the status field is defined as a subresouce of the flyteworkflow CRD. This is
+// Check to see if the status field is defined as a subresource of the flyteworkflow CRD. This is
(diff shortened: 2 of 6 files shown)


hamersaw:
@EngHabu is it sane to structure the hasStatusSubresourceAPI handling in the controller logic? i'm trying to convince myself this approach is better than burying it in the workflowstore Update function. is there a compelling reason why workflowstore has separate Update and UpdateStatus functions besides mimicing the kubernetes API? logically, the controller shouldn't care what API calls update the flyteworkflow CRD, just that it happens correctly.

hamersaw:
It was decided that at the moment support for this feature faces complications in deployment and upgrading of flytepropeller and flyte kubernetes components. See https://github.com/flyteorg/flyte/issues/126 for further explanation.

PR #343, opened 2021-10-19
Vault Secret Manager #patch

bimtauer:
# TL;DR
This introduces a new SecretManager for injecting Vault secrets from an existing deployment of Vault Agent. 

The new injector takes flyte secret requests and parses them into the necessary annotations to trigger Vault Agent's webhook and mount the secrets to a task pod. It also configures the necessary env vars to let the pod know where to find the secrets. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [] Any pending items have an associated Issue

## Complete description
The VaultSecretManager can be configured by setting: 
```yaml
secretManagerType: Vault
vaultSecretManager:
  role: <a-valid-vault-role>
```
When calling its `Inject` method, the VaultSecretManager will first add the common Vault annotations, that is:
```go
"vault.hashicorp.com/agent-inject":            "true",
"vault.hashicorp.com/secret-volume-path":      filepath.Join(VaultSecretPathPrefix...),
"vault.hashicorp.com/role":                    i.cfg.Role,
"vault.hashicorp.com/agent-pre-populate-only": "true",
```
_Note: we only run the Vault Agent init container to inject the secrets once since the Vault Agent sidecar connector capable of refreshing secrets would otherwise prevent the pod from completing. Very long running tasks could in theory exceed the TTL of a vault secret depending on configuration._ 

Then for each secret, it will configure matching annotations that will result in the requested secret being mounted at:
`/etc/flyte/secrets/SECRET_GROUP/SECRET_NAME` through adding the following annotations: 
```go
fmt.Sprintf("vault.hashicorp.com/agent-inject-secret-%s", id): secret.Group,
fmt.Sprintf("vault.hashicorp.com/agent-inject-file-%s", id):   fmt.Sprintf("%s/%s", secret.Group, secret.Key),
fmt.Sprintf("vault.hashicorp.com/agent-inject-template-%s", id): fmt.Sprintf(`
  {{- with secret "%s" -}}
  {{ .Data.data.%s }}
  {{- end -}}`, secret.Group, secret.Key),
```
The annotations for a single secret are grouped using a UUID called `id` in the above snippet. 
A design choice here was to interpret the SECRET_GROUP as a vault secret path and the SECRET_NAME as the key to a single secret value of multiple possible k:v mappings in a vault secret. 

In python you would thus request a Vault secret called "username" from a Vault secret file called "internal/data/database/config" in the following manner: 
```python
@task(secret_requests=[Secret(group="internal/data/database/config", key="username"))
```

## Tracking Issue
_NA_

## Follow-up issue
_NA_ - Need to update docs to explain usage and how this requires one to add and configure Vault Agent on one's cluster.

### DIFF
--- a/pkg/webhook/utils.go
+++ b/pkg/webhook/utils.go
@@ -120,6 +120,10 @@ func AppendVolume(volumes []corev1.Volume, volume corev1.Volume) []corev1.Volume
 }
 
 func CreateVaultAnnotationsForSecret(secret *core.Secret, kvversion config.KVVersion) (map[string]string, error) {
+	// Creates three grouped annotations "agent-inject-secret", "agent-inject-file" and "agent-inject-template"
+	// for a given secret request and KV engine version. The annotations respectively handle: 1. retrieving the
+	// secret from a vault path specified in secret.Group, 2. storing it in a file named after secret.Group/secret.Key
+	// and 3. creating a template that retrieves only secret.Key from the multiple k:v pairs present in a vault secret.
 	id := string(uuid.NewUUID())
 
 	// Set the consul template language query depending on the KV Secrets Engine version.


kumare3:
Very small nits from me. Fantastic work @bimtauer. thank you. +1, please address the nits

PR #351, opened 2021-10-21
Scale out with propeller manager and workflow sharding

hamersaw:
# TL;DR
Adding a FlytePropeller Manager component which is responsible for configuring and ensuring liveness over a collection of FlytePropeller instances. FlyteWorkflow CRDs are effectively sharded over these instances, so that there is a deterministic one-to-one relationship. this enables horizontally scaling FlytePropeller.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Functionality has been implemented in accordance to the accepted [RFC](https://github.com/flyteorg/flyte/blob/master/rfc/system/1483-flytepropeller-horizontal-scaling.md).

## Tracking Issue
https://github.com/flyteorg/flyte/issues/125

## Follow-up issue
_NA_


cmd/controller/cmd/root.go
```diff
@@ -144,11 +144,30 @@ func getKubeConfig(_ context.Context, cfg *config2.Config) (*kubernetes.Clientse
...
+	}
+
 	opts := []informers.SharedInformerOption{
 		informers.WithTweakListOptions(func(options *v1.ListOptions) {
-			options.LabelSelector = v1.FormatLabelSelector(controller.IgnoreCompletedWorkflowsLabelSelector())
+			options.LabelSelector = v1.FormatLabelSelector(labelSelector)
 		}),
 	}
+
+	/*opts := []informers.SharedInformerOption{
```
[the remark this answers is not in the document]
hamersaw: done

cmd/controller/cmd/root.go
```diff
@@ -144,11 +144,30 @@ func getKubeConfig(_ context.Context, cfg *config2.Config) (*kubernetes.Clientse
...
 func sharedInformerOptions(cfg *config2.Config) []informers.SharedInformerOption {
+	labelSelector := controller.IgnoreCompletedWorkflowsLabelSelector()
+	if len(cfg.IncludeShardKey) > 0 {
+		labelSelectorRequirement := v1.LabelSelectorRequirement{"shardKey", v1.LabelSelectorOpIn, cfg.IncludeShardKey}
+
+		labelSelector.MatchExpressions = append(labelSelector.MatchExpressions, labelSelectorRequirement)
+	}
+
+	if len(cfg.ExcludeShardKey) > 0 {
+		labelSelectorRequirement := v1.LabelSelectorRequirement{"shardKey", v1.LabelSelectorOpNotIn, cfg.IncludeShardKey}
```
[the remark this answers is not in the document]
hamersaw: exclude is being used for the "enableUncoveredReplica" option. mainly for the project and domain shard strategies, it's only included on the shard key option for completeness - but it's easy to remove.

cmd/manager/cmd/root.go
```diff
@@ -0,0 +1,195 @@
...
+	"flag"
+	"os"
+	"runtime"
+	//"runtime/pprof" // TODO hamersaw - add pprof
+	"strings"
+
+	"github.com/flyteorg/flytestdlib/config"
+	"github.com/flyteorg/flytestdlib/config/viper"
+	"github.com/flyteorg/flytestdlib/logger"
+	// TODO hamersaw - enable
```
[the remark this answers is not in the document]
hamersaw: done

cmd/manager/cmd/root.go
```diff
@@ -0,0 +1,195 @@
...
+	kubecfg.Timeout = cfg.KubeConfig.Timeout.Duration
+
+	kubeClient, err := kubernetes.NewForConfig(kubecfg)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "Error building kubernetes clientset")
+	}
+	return kubeClient, kubecfg, err
+}
+
+func safeMetricName(original string) string {
```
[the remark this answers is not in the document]
hamersaw: done

cmd/manager/cmd/root.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return nil
+}
+
+func logAndExit(err error) {
+	logger.Error(context.Background(), err)
+	os.Exit(-1)
+}
+
+func getKubeConfig(_ context.Context, cfg *propellerConfig.Config) (*kubernetes.Clientset, *restclient.Config, error) {
```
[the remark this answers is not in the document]
hamersaw: done

manager/config/config.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/flyteorg/flytestdlib/config"
+)
+
+//go:generate pflags Config --default-var=DefaultConfig
+
+var (
+	DefaultConfig = &Config{
+		PodApplication: "flytepropeller",
+		PodNamespace: "flyte",
+		PodTemplate: "flytepropeller-template",
```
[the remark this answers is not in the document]
hamersaw: i sure think so!

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/flyteorg/flytestdlib/logger"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type Manager struct {
```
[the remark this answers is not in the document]
hamersaw: done

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+// TODO hamersaw - integrate prometheus metrics
+/*type metrics struct {
+	Scope            promutils.Scope
+	EnqueueCountWf   prometheus.Counter
+	EnqueueCountTask prometheus.Counter
+}*/
+
+func (m *Manager) recoverPods(ctx context.Context) error {
+	// TODO hamersaw - do we need to handle pods with Error status?
+	// with 3 pods locally we get "too many open files"
```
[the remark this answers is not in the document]
hamersaw: changed to createPods (to complement the deletePods function)

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type Manager struct {
+	kubePodsClient corev1.PodInterface
+	pod            *v1.Pod
+	podApplication string
+	podNames       []string
```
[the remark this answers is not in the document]
hamersaw: done

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	for _, podName := range m.podNames {
+		podExists[podName] = false
+	}
+
+	for _, pod := range pods.Items {
+		podName := pod.ObjectMeta.Name
+		if _, ok := podExists[podName]; ok {
+			podExists[podName] = true
+		} else {
+			logger.Warnf(ctx, "detected unmanaged pod '%s'", podName)
```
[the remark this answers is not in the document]
hamersaw: metric added. i have added logging to show pods that aren't currently being managed. the only issue i see is pods that are stale between manager executions (ie. started by instance 1, instace 1 is shutdown, and instance 2 sees their pods). again, this shouldn't happen, and will be present as an error in the log. but it's a big question of how defensive we need to be.

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+
+			logger.Infof(ctx, "created pod '%s'", podName)
+		}
+	}
+
+	return nil
+}
+
+func (m *Manager) Run(ctx context.Context) error {
```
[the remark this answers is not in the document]
hamersaw: done

manager/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var podNames []string
+	for i := 0; i < podCount; i++ {
+		podNames = append(podNames, fmt.Sprintf("%s-%d", cfg.PodApplication, i))
+	}
+
+	manager := &Manager{
```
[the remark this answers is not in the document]
hamersaw: done

cmd/controller/cmd/root.go
```diff
@@ -116,45 +112,39 @@ func logAndExit(err error) {
...
-	kubeClient, err := kubernetes.NewForConfig(kubecfg)
-	if err != nil {
-		return nil, nil, errors.Wrapf(err, "Error building kubernetes clientset")
-	}
-	return kubeClient, kubecfg, err
-}
-
 func sharedInformerOptions(cfg *config2.Config) []informers.SharedInformerOption {
+	labelSelector := controller.IgnoreCompletedWorkflowsLabelSelector()
+	addLabelSelector(labelSelector, transformers.ShardKeyLabel, v1.LabelSelectorOpIn, cfg.IncludeShardKeyLabel)
```
kumare3, line 117: how is this supposed to work?
So if we want to run for all - what do we do? Should these not be behind if conditions?

manager/config/config.go
```diff
@@ -0,0 +1,55 @@
...
+
+type ShardType = string
+
+const (
+	ConsistentHashingShardType ShardType = "consistent-hashing"
+	// TODO - implement
+	//ProjectShardType           ShardType = "project"
+)
+
+type ShardConfig struct {
```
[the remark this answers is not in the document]
kumare3: ping, can we add a doc for this config of the format

// ShardConfig specifies ....

manager/shard_strategy.go
```diff
@@ -0,0 +1,191 @@
...
+	} else {
+		return fmt.Errorf("invalid podIndex '%d' out of range [0,%d)", podIndex, h.GetPodCount())
+	}
+
+	return nil
+}
+
+// Computes a [startKey, endKey) pair denoting the key responsibilities for the provided pod index
+// given the keyspaceSize and podCount parameters
+func computeKeyRange(keyspaceSize, podCount, podIndex int) (int, int) {
```
kumare3, line 99: I think this should be an exported method or util method and have tests on it

manager/shard_strategy.go
```diff
@@ -0,0 +1,191 @@
...
+	keyRemainder := keyspaceSize - (podCount * keysPerPod)
+
+	return computeStartKey(keysPerPod, keyRemainder, podIndex), computeStartKey(keysPerPod, keyRemainder, podIndex+1)
+}
+
+func computeStartKey(keysPerPod, keysRemainder, podIndex int) int {
+	return (intMin(podIndex, keysRemainder) * (keysPerPod + 1)) + (intMax(0, podIndex-keysRemainder) * keysPerPod)
+}
+
+func intMin(a, b int) int {
```
kumare3, line 110: i know sad, we should just write these in flytestdlib one day

manager/shard_strategy.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return nil
+}
+
+func getFlytePropellerContainer(pod *v1.PodSpec) (*v1.Container, error) {
+	// find flytepropeller container(s)
+	var containers []*v1.Container
+	for i := 0; i < len(pod.Containers); i++ {
+		commands := pod.Containers[i].Command
+		if len(commands) > 0 && commands[0] == "flytepropeller" {
```
kumare3, line 181: this seems odd, its ok to probably use an annotation or the name itself of the container?

manager/manager.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+		},
+		m.scanInterval,
+	)
+
+	logger.Infof(ctx, "shutting down manager")
+	return nil
+}
+
+func (m *Manager) shutdown() {
```
kumare3, line 196: why delete pods on shutdown?

kumare3 (approved):

manager/manager.go
kumare3: also how is the shutdown invoked? We dont want the pods to go down if manager gets a SIGINT

kumare3:
This mostly looks good except for shutdown.
I think we could add a new `/http` handle that can have a `/shutdown` and `/restart` calls. But shutting down otherwise seems dangerous

manager/shard_strategy.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/flyteorg/flytepropeller/manager/config"
+	"github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+type ShardStrategy interface {
+	GetPodCount() int
+	UpdatePodSpec(pod *v1.PodSpec, podIndex int) error
```
kumare3, line 15:
```suggestion
	UpdatePodSpec(pod *v1.PodSpec, podIndex int, maxPods int) error
```

What if we pass the maxPods as an input, this way, the shardStrategy is independent of the scale. This will allow us to - if we want in the future - create an auto-scaling flytepropeller?

kumare3: I guess you think - it will not work with per-namespace. but what if that is always `1` for per namespace and if we ever want to increase propellers per namespace we can?

manager/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type ReplicaConfig struct {
+	Entities []string `json:"entities" pflag:",The list of entities to be managed"`
+}
+
+type ShardConfig struct {
+	Type                   ShardType       `json:"type" pflag:"\"hash\",Shard implementation to use"`
+	EnableUncoveredReplica bool            `json:"enable-uncovered-replica" pflag:"\"true\",Start a replica to process FlyteWorkflows that are uncovered by the configured shard label selectors"`
+	Replicas               []ReplicaConfig `json:"replicas" pflag:"-"`
```
kumare3, line 45: what is difference between replicas and podCount?

isnt podCount like the replica?

manager/shard_strategy.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return b
+}
+
+// The ProjectShardStrategy assigns project(s) to individual FlytePropeller instances to
+// determine FlyteWorkflow processing responsibility.
+type EnvironmentShardStrategy struct {
+	envType                environmentType
+	enableUncoveredReplica bool
+	replicas               [][]string
```
kumare3, line 131: can we define replicas please? I do not follow

manager/shard_strategy.go
hamersaw: +1

manager/config/config.go
hamersaw: podCount is used with hashing, replicas (may be terribly named) but refers to a collection of identifiers (e.x. projects / domains) for use with those sharding mechanisms ... do we add separate fields for ProjectReplicas and DomainReplicas? I tried to adhere to the current config setups, but perhaps this case doesn't fit well?

manager/shard_strategy.go
hamersaw: maybe should be named identifiers? the idea is we initialize a number of replicas each consisting of a collection of ids for project and domain sharding. For example:
replica 1 is responsible for projects "flytesnacks" and "flyteexamples"
replica 2 is responsible for "foo"
and then if enableUncoveredReplica is set true then a 3rd replica is started using negations so it handles all other projects

cmd/controller/cmd/root.go
hamersaw: the addLabelSelector function has a conditional inside checking if the shard label keys contains anything. change the name to addLabelSelectorIfExists?

manager/manager.go
hamersaw: if manager 1 starts and uses hash sharding with 4 replicas we have a 32 bit range distributed with 8 for each, the manager fails and flytepropeller-0, flytepropeller-1, 2, and 3 ... are still running. if another managers starts sharding with 8 replicas than it starts flytepropeller-4, 5, 6, 7. however, flytepropeller-3 covers a superset of keys that 4 does similar overlap issues with project and domain sharding. i figured any shutdown on the manager kills all underlying replicas - and that will keep restarts / autoscaling / etc much more clean. maybe something we need to talk about?

manager/config/config.go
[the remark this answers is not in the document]
hamersaw: done

cmd/controller/cmd/root.go
hamersaw: i changed the function name. we could otherwise define a labelSelector struct like we do in many of the tests to remove the addLabelSeletor function?

```
type labelSelector struct {
    label string
    operation v1.LabelSelectorOperator
    values []string
}

labelSelectors := []labelSelector{
    labelSelector{transformers.ShardKeyLabel, v1.LabelSelectorOpIn, cfg.IncludeShardKeyLabel},
    labelSelector{transformers.ShardKeyLabel, v1.LabelSelectorOpNot, cfg.ExcludeShardKeyLabel},
    ...
}

for _, labelSelector := range labelSelectors {
    if len(labelSelector.values) > 0 {
        // add label selector
    }
}
```

manager/shard_strategy.go
hamersaw: I exported it. The testing should be very thorough as is, we're testing for increasing podCounts up to a max test keyspaceSize of 32. Each test ensures that the lowest and highest number of keys assigned to a specific pod doesn't vary by more than 1.

I'll add a note to refactor this out to a util method when we refactor the project structure.

manager/manager.go
hamersaw: fixed this per our discussion. we are not storing the pod template resource version and shard strategy hash within each pod, if either changes we delete the pod and restart with the new information.

manager/shard_strategy.go
hamersaw: updated to use a configurable container name. much cleaner idea.

manager/shard_strategy.go
hamersaw: i'm having some trouble understanding this completely. as you mentioned in the case of hashing the shardStrategy can easily be independent of the scale, but it is very difficult to fit into the project and domain shard strategies - which are basically defined by scale.

my idea for supporting autoscaling just requires updating the ShardStrategy (so going the opposite direction of making them independent). for the case of HashShardStrategy if we increase / decrease the podCount it will change the hash value and consequently restart the necessary pods. it should be a really simple addition to the createPods loop that we currently have, we can add simple scale up and down functions onto the interface and in the loop something like:

```
if scaleUp {
    shardStrategy.ScaleUp()
} else if scaleDown {
    shardStrategy.ScaleDown()
}
```

as i mentioned, downstream the loop identifies the shardStrategy changes and will automatically update the running pods.

kumare3: ok

manager/manager.go
```diff
@@ -0,0 +1,295 @@
+package manager
```
kumare3, line 1: Lets add package level docs here
```suggestion
// package manager contains ....
package manager
```

These show up in a special way in the generated docs. Also we should write things like
this one - https://github.com/flyteorg/flyteadmin/blob/master/scheduler/doc.go

kumare3: I want to understand the assumptions, the high level design and what are unhandled cases

manager/shard_strategy.go
```diff
@@ -0,0 +1,226 @@
...
+	"fmt"
+	"hash/fnv"
+
+	"github.com/flyteorg/flytepropeller/manager/config"
+	"github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// Defines necessary functionality for a sharding strategy
```
kumare3, line 16:
```suggestion
// ShardStrategy defines necessary functionality for a sharding strategy
```
the style is to write the name of the struct followed by the sentence. Golint will complain otherwise

manager/shard_strategy.go
```diff
@@ -0,0 +1,226 @@
...
+type ShardStrategy interface {
+	// Returns the total number of pods for the sharding strategy
+	GetPodCount() int
+	// Generates a unique hash code to identify shard strategy updates
+	HashCode() (uint32, error)
+	// Updates the PodSpec for the specified index to include label selectors
+	UpdatePodSpec(pod *v1.PodSpec, containerName string, podIndex int) error
+}
+
+// Creates and validates a new ShardStrategy defined by the configuration
```
kumare3, line 26:
```suggestion
// NewShardStrategy creates and validates a new ShardStrategy defined by the configuration
```

same here. Also repeat this for all docs. Lets make propeller even better documented

manager/shard_strategy.go
```diff
@@ -0,0 +1,226 @@
...
+		}, nil
+	}
+
+	return nil, fmt.Errorf("shard strategy '%s' does not exist", shardConfig.Type)
+}
+
+// HashShardStrategy evenly assigns disjoint keyspace responsibilities over a collection of pods.
+// All FlyteWorkflows are assigned a shard-key using a hash of their executionID and are then
+// processed by the FlytePropeller instance responsible for that keyspace range.
+type HashShardStrategy struct {
```
kumare3, line 71: can we split each of the strategies into separate files?
One way would be to create a folder called `shard_strategy` and then under this folder
hash.go
env.go
etc?

manager/shard_strategy.go
```diff
@@ -0,0 +1,226 @@
...
+	EnableUncoveredReplica bool
```
kumare3, line 72: what does this mean?

kumare3: ohh do you mean a hot-standby?

manager/shard_strategy.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	return b
+}
+
+// The ProjectShardStrategy assigns project(s) to individual FlytePropeller instances to
+// determine FlyteWorkflow processing responsibility.
+type EnvironmentShardStrategy struct {
+	EnvType                environmentType
+	EnableUncoveredReplica bool
```
kumare3, line 143: how does this really work? i thought we are assigning bins to specific propellers right?

manager/shard_strategy.go
hamersaw: done

manager/shard_strategy.go
hamersaw: resolved per our discussion

manager/shard_strategy.go
hamersaw: resolved per our discussion

kumare3:
@hamersaw I was talking with @EngHabu and he mentioned one important thing. You should create all the pods, with an owner-reference. The owner reference should be the same as - the flytepropeller deployment owner reference that is populated on the manager.
Thus the manager should pass down its own owner-reference to the propeller pods.

manager/manager.go
hamersaw: done

hamersaw:
> @hamersaw I was talking with @EngHabu and he mentioned one important thing. You should create all the pods, with an owner-reference. The owner reference should be the same as - the flytepropeller deployment owner reference that is populated on the manager. Thus the manager should pass down its own owner-reference to the propeller pods.
> 
> This will enable that when you delete the deployment it will automatically delete all the pods that were created

This is completed, to track this we added the owner references (if they exist) from the pod which the manager is started in to the manager pods.

This approach differs from the previous shutdown hook which deleted all managed pods when the manager exited. Meaning a failure in the manager would delete them as well as a successful shutdown. In our new approach, failures in the manager pod do not affect managed pods.

pkg/utils/k8s.go
```diff
@@ -84,6 +96,49 @@ func ToK8sResourceRequirements(resources *core.Resources) (*v1.ResourceRequireme
 	return res, nil
 }
 
+func GetContainer(pod *v1.PodSpec, containerName string) (*v1.Container, error) {
+	// find flytepropeller container(s)
+	var containers []*v1.Container
+	for i := 0; i < len(pod.Containers); i++ {
+		if pod.Containers[i].Name == containerName {
```
EngHabu, line 103: I don't think a Pod spec with multiple containers having the same name is valid. Unless the validation on PodTemplate is more lax?

EngHabu (commented):
A couple of nits

cmd/manager/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"github.com/flyteorg/flytepropeller/cmd/manager/cmd"
+)
+
+func init() {
+}
```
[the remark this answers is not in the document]
hamersaw: done

pkg/utils/k8s.go
hamersaw: done. seems this was a remnant of checking the pod command instead of the name.

manager/manager.go
```diff
@@ -0,0 +1,290 @@
...
+					Labels:          podLabels,
+					OwnerReferences: m.ownerReferences,
+				},
+				Spec: *podTemplate.Template.Spec.DeepCopy(),
+			}
+
+			err := m.shardStrategy.UpdatePodSpec(&pod.Spec, m.podTemplateContainerName, i)
+			if err != nil {
+				logger.Errorf(ctx, "failed to update pod spec for '%s' [%v]", podName, err)
+				continue
```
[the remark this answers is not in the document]
hamersaw: much better! i'll be sure to remember this module.

manager/manager.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+
+		// update podExists to track existing pods
+		if _, ok := podExists[podName]; ok {
+			podExists[podName] = true
+
+			if pod.Status.Phase == v1.PodRunning {
+				podsRunning++
+			} else if pod.Status.Phase == v1.PodFailed {
+				logger.Warnf(ctx, "flytepropeller pod '%s' in 'failed' state", podName)
```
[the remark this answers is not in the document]
hamersaw: updated to delete to pod. it was just a warning to make debugging an issue easier. i agree deleting and restarting as you suggested is a better idea.

manager/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	DefaultConfig = &Config{
+		PodApplication:           "flytepropeller",
+		PodTemplateContainerName: "flytepropeller",
+		PodTemplateName:          "flytepropeller-template",
+		PodTemplateNamespace:     "flyte",
+		ScanInterval: config.Duration{
+			Duration: 10 * time.Second,
+		},
+		ShardConfig: ShardConfig{
+			Type:       "hash",
```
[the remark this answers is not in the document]
hamersaw: done

manager/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+// Configuration for defining shard replicas when using project or domain shard types
+type PerShardMappingsConfig struct {
+	IDs []string `json:"ids" pflag:",The list of ids to be managed"`
+}
+
+// Configuration for the FlytePropeller sharding strategy
+type ShardConfig struct {
+	Type             ShardType                `json:"type" pflag:"\"hash\",Shard implementation to use"`
+	PerShardMappings []PerShardMappingsConfig `json:"per-shard-mapping" pflag:"-"`
+	ShardCount       int                      `json:"shard-count" pflag:"\"3\",The number of shards to manage for a 'hash' shard type"`
```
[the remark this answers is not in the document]
hamersaw: done

manager/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	configSection = config.MustRegisterSection("manager", DefaultConfig)
+)
+
+type ShardType = string
+
+const (
+	DomainShardType  ShardType = "domain"
+	ProjectShardType ShardType = "project"
+	HashShardType    ShardType = "hash"
+)
```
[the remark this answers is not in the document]
hamersaw: @EngHabu i'm not sure i fully understand this. we're using these consts to define the type in the config similar to the [queue configuration](https://github.com/flyteorg/flytepropeller/blob/7df7c0a8fde25e65fa709fe4e8e02062365850f3/pkg/controller/config/config.go#L154). if we switch this to an integer then we can't use the ShardType type in the Config. instead we would use a plain string and convert it to this enum correct? it seems there is some support for this within the generated enum, but it gets a little messy.

also, the string conversion means the configuration name needs to match the name of these consts exactly (i.e. 'type: "ShardTypeDomain"', 'type: "ShardTypeProject";, etc) unless we change these names to Domain, Project, Hash and generate them in snakecase with the enumer flag --transform=snake_case. then 'type: "domain"', 'type: "hash"', etc would work.

is there a much easier way to do this?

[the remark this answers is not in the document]
kumare3: https://github.com/flyteorg/flytepropeller/blob/16b0d2c12625e013eb1a7c479f8cfa062075501e/pkg/webhook/config/config.go#L9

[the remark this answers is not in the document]
hamersaw: done. much gooder, i think the -yaml flag allows automatic parsing with the decode yaml function? and somehow the trimprefix flag on enumer eluded me.

manager/config/config.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Configuration for defining shard replicas when using project or domain shard types
+type PerShardMappingsConfig struct {
+	IDs []string `json:"ids" pflag:",The list of ids to be managed"`
+}
+
+// Configuration for the FlytePropeller sharding strategy
+type ShardConfig struct {
+	Type             ShardType                `json:"type" pflag:"-,Shard implementation to use"`
```
EngHabu, line 45: If you have the latest pflags binary, you should be able to generate binding for this... does it error?

```suggestion
	Type             ShardType                `json:"type" pflag:",Shard implementation to use"`
```

kumare3 (approved):

manager/config/config.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Configuration for the FlytePropeller Manager instance
+type Config struct {
+	PodApplication           string          `json:"pod-application" pflag:"\"flytepropeller\",Application name for managed pods"`
+	PodTemplateContainerName string          `json:"pod-template-container-name" pflag:"\"flytepropeller\",The container name within the K8s PodTemplate name used to set FlyteWorkflow CRD labels selectors"`
+	PodTemplateName          string          `json:"pod-template-name" pflag:"\"flytepropeller-template\",K8s PodTemplate name to use for starting FlytePropeller pods"`
+	PodTemplateNamespace     string          `json:"pod-template-namespace" pflag:"\"flyte\",Namespace where the k8s PodTemplate is located"`
+	ScanInterval             config.Duration `json:"scan-interval" pflag:"\"10s\",Frequency to scan FlytePropeller pods and start / restart if necessary"`
+	ShardConfig              ShardConfig     `json:"shard" pflag:",Configure the shard strategy for this manager"`
```
EngHabu, line 57: Let's drop all the "default" values from here, this is ignored anyway when you have `--default-var=DefaultConfig` above and makes this code a bit harder to read...

```suggestion
	PodApplication           string          `json:"pod-application" pflag:",Application name for managed pods"`
	PodTemplateContainerName string          `json:"pod-template-container-name" pflag:",The container name within the K8s PodTemplate name used to set FlyteWorkflow CRD labels selectors"`
	PodTemplateName          string          `json:"pod-template-name" pflag:",K8s PodTemplate name to use for starting FlytePropeller pods"`
	PodTemplateNamespace     string          `json:"pod-template-namespace" pflag:",Namespace where the k8s PodTemplate is located"`
	ScanInterval             config.Duration `json:"scan-interval" pflag:",Frequency to scan FlytePropeller pods and start / restart if necessary"`
	ShardConfig              ShardConfig     `json:"shard" pflag:",Configure the shard strategy for this manager"`
```

EngHabu (approved):
Feel free to ignore...

manager/config/config.go
hamersaw: done

manager/config/config.go
hamersaw: done

EngHabu (approved):

PR #370, opened 2021-12-02
Kenny/type metadata

kennyworkman:
## Strip Annotations from TaskTemplate#minor

Strips any annotation from type so runtime task structure remains the same size

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

re:
* https://github.com/flyteorg/flytekit/pull/759
* https://github.com/flyteorg/flyteidl/pull/232


pkg/compiler/transformers/k8s/node.go
```diff
@@ -221,6 +221,18 @@ func buildTasks(tasks []*core.CompiledTask, errs errors.CompileErrors) map[commo
 				errs.Collect(errors.NewValueCollisionError(taskID, "Id", taskID))
 			}
 
+			// We don't want Annotation data available at task runtime for performance.
+			if flyteTask.Template != nil &&
```
kennyworkman, line 225: i'm going to be honest this is pretty gross, but don't know the most semantic way of dereferencing nested pointers here given that the whole struct is received as a big chunk from admin

pkg/compiler/transformers/k8s/node.go
```diff
@@ -221,6 +221,18 @@ func buildTasks(tasks []*core.CompiledTask, errs errors.CompileErrors) map[commo
...
+				flyteTask.Template.Interface != nil &&
+				flyteTask.Template.Interface.Inputs != nil &&
+				flyteTask.Template.Interface.Inputs.Variables != nil {
+				for _, v := range flyteTask.Template.Interface.Inputs.Variables {
```
EngHabu, line 229: We should probably do the same with outputs for consistency/completeness sake....

EngHabu (commented):
One more change that needs to happen in the compiler is the type check (that checks whether a given output can be passed down as a given input type).. this check relies on the [serialized representation of LiteralType here](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/validators/typing.go#L55) and having annotations in the output but not the input (or vis versa) will fail such validation I believe...

Just list we "nil" the metadata right before that line, we should nil the Annotations (just for checking)...

pkg/compiler/transformers/k8s/node.go
kennyworkman: addressed with new commit

kennyworkman:
@EngHabu not sure how you can verify the tests pass with github actions alone because the flyteidl source code structure doesnt play well with go mod

i just did this locally to confirm that everything passes with new flyteidl
`replace github.com/flyteorg/flyteidl => /Users/kenny/latch/flyteorg/flyteidl`

EngHabu:
> @EngHabu not sure how you can verify the tests pass with github actions alone because the flyteidl source code structure doesnt play well with go mod
> 
> 
> 
> i just did this locally to confirm that everything passes with new flyteidl
> 
> `replace github.com/flyteorg/flyteidl => /Users/kenny/latch/flyteorg/flyteidl`

You can do this instead:
```golang
replace github.com/flyteorg/flyteidl => github.com/<fork>flyteidl <Sha>
```

And this should work fine in CI

kennyworkman:
@EngHabu good to go

pkg/compiler/transformers/k8s/node.go
```diff
@@ -221,6 +221,30 @@ func buildTasks(tasks []*core.CompiledTask, errs errors.CompileErrors) map[commo
 				errs.Collect(errors.NewValueCollisionError(taskID, "Id", taskID))
 			}
 
+			// We don't want Annotation data available at task runtime for performance.
+			if flyteTask.Template != nil &&
+				flyteTask.Template.Interface != nil {
+
+				if flyteTask.Template.Interface.Inputs != nil &&
+					flyteTask.Template.Interface.Inputs.Variables != nil {
```
EngHabu, line 229: If it makes you feel any better you can make this:
```suggestion
				if flyteTask.Template.Interface.Inputs != nil {
```

Since `range nil` will yield 0 iterations and is an accepted value by go... but 🤷🏽

EngHabu (commented):
I don't see [this](https://github.com/flyteorg/flytepropeller/pull/370#pullrequestreview-822917601) addressed in the PR... please let me know if you want to discuss further

go.mod
```diff
@@ -33,3 +33,4 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
+replace github.com/flyteorg/flyteidl => github.com/kennyworkman/flyteidl 17fc33f66a9d32c22f93adbb30166857fbb6bc11
```
EngHabu, line 36: You will also need to run `go mod tidy` to update the `go.sum` file...

EngHabu:
Mind deleting `pkg/compiler/transformers/k8s/.node_test.go.swp` from the PR as well?

go.mod
```diff
@@ -33,3 +34,5 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
+
+replace github.com/flyteorg/flyteidl => github.com/kennyworkman/flyteidl v0.21.12-0.20211210005601-8c4bb8cda5fb
```
EngHabu, line 38: just a reminder to update when idl PR merges..

EngHabu (approved):
Looks good to me!

go.mod
kennyworkman: do you want me to resolve conflict and merge with my version of idl?

eapolinario (approved):

EngHabu (approved):

PR #376, opened 2021-12-10
Binding Panic when SDK produces an empty binding

EngHabu:

# TL;DR
Fix a panic (and improve the error) when SDK produces an empty binding

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1935

### DIFF
--- a/pkg/compiler/validators/bindings.go
+++ b/pkg/compiler/validators/bindings.go
@@ -17,6 +17,11 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 
 	switch val := binding.GetValue().(type) {
 	case *flyte.BindingData_Collection:
+		if val.Collection == nil {
+			errs.Collect(errors.NewParameterNotBoundErr(nodeID, nodeParam))
+			return nil, nil, !errs.HasErrors()
+		}
+
 		if expectedType.GetCollectionType() != nil {
 			allNodeIds := make([]c.NodeID, 0, len(val.Collection.GetBindings()))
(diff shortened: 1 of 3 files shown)


pkg/compiler/validators/bindings.go
```diff
@@ -15,12 +15,12 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 	expectedType *flyte.LiteralType, errs errors.CompileErrors) (
 	resolvedType *flyte.LiteralType, upstreamNodes []c.NodeID, ok bool) {
 
-	switch binding.GetValue().(type) {
+	switch val := binding.GetValue().(type) {
 	case *flyte.BindingData_Collection:
 		if expectedType.GetCollectionType() != nil {
-			allNodeIds := make([]c.NodeID, 0, len(binding.GetMap().GetBindings()))
+			allNodeIds := make([]c.NodeID, 0, len(val.Collection.GetBindings()))
```
kumare3, line 21: i think we should check `if val.GetCollection() != nil`

EngHabu: I can add it... I don't think it adds value though... these objects are parsed protos. If this is pure go code, I would say we should check because val.Collection may have the interface binding but not the value set. Added for being extra defensive :-)

hamersaw (approved):

PR #385, opened 2022-01-13
Add templating functionality to AddObjectMetadata

bimtauer:
# TL;DR
Implements https://github.com/flyteorg/flyte/issues/1592 for `DefaultAnnotations` and `DefaultLabels`

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Adds a new function `InjectTemplates` which is called inside `AddObjectMetadata`. 
The functions retrieves the current `project` and `domain` by traversing the `TaskExecutionMetadata`. It then first marshalls the objects selected for templating into yaml, performs string replacement of the placeholders and unmarshalls that yaml back. Admittedly slightly hacky, would be curious to learn a better way!

Tests are added, smoke testing was already done successfully on our Flyte dev cluster.  

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1592

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/io"
 	"github.com/flyteorg/flytestdlib/contextutils"
 	stdErrors "github.com/flyteorg/flytestdlib/errors"
+	"github.com/ghodss/yaml"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -49,7 +50,9 @@ import (
 )
 
 const finalizer = "flyte/flytek8s"
-
+const templateProjectVariable = "{{ PROJECT }}"
+const templateDomainVariable = "{{ DOMAIN }}"
+const replaceAllInstancesOfString = -1
 const pluginStateVersion = 1
 
 type PluginPhase uint8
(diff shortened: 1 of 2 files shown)


pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -103,10 +106,37 @@ type PluginManager struct {
...
 
+// Handles dynamic templating of DefaultAnnotations and DefaultLabels to replace placeholders {{ PROJECT }} and {{ DOMAIN }}
+// with there actual values at execution time. This is useful if one wants to add custom IAM roles or Vault roles to task pods
+// depending on the project and domain.
+func InjectTemplates(taskCtx pluginsCore.TaskExecutionMetadata, defaults map[string]string) map[string]string {
+	execId := taskCtx.GetTaskExecutionID().GetID().NodeExecutionId.GetExecutionId()
+	project := execId.GetProject()
+	domain := execId.GetDomain()
+
+	y, err := yaml.Marshal(defaults)
```
EngHabu, line 117: I would avoid marshaling and unmarshaling (even though admittedly it may make the code look cleaner)...

This code has been [done here](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/template/template.go) already... let's just use that instead..

pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -49,7 +50,9 @@ import (
 )
 
 const finalizer = "flyte/flytek8s"
-
+const templateProjectVariable = "{{ PROJECT }}"
+const templateDomainVariable = "{{ DOMAIN }}"
```
EngHabu, line 54: I think admin [here](https://github.com/flyteorg/flyteadmin/blob/76be8471380dbb6e58eeea4a6feb1601812eb9fb/pkg/clusterresource/controller.go#L61-L64) uses lower case to represent these variables... we should either keep this consistent or case insensitive.

I think you should leverage this [package instead](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/template/template.go) and just add project and domain as other fields in the Parameters struct... make them available to other templating we do throughout.

bimtauer: Alright, thanks for the hint, wasn't aware of that package. I chose UPPERCASE to signalize that these are variables that will be filled by flyte itself unlike those in admin which I believe get filled with user-defined values that could have arbitrary keys. Can change to lowercase if you want though.

bimtauer: I have the basic outline here https://github.com/bimtauer/flyteplugins/commit/9b6026ea6c47c6e9c7213302858537038572b64a but it looks like `Parameters` is very much tied to that file/s3 path use case. In this use-case now, all that is needed is passing the `TaskExecutionMetadata` to retrieve Project and Domain. But the other fields are not nullable. What to do here?

bimtauer: To clarify: I'd like to be able to call `Render` with params simply being: 
```go
Parameters{
    TaskExecMetadata: taskMetadata,
}
```

bimtauer: I removed marshalling now in favor of directly accessing the k,v pairs. Let me know what you think.

PR #399, opened 2022-02-11
Refactor for single binary

kumare3:
# TL;DR
This PR moves some of the entrypoint logic into the controller package. This makes the controller invokable externally as part of the single binary push

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added


hamersaw (commented):
Will defer on the initCerts question. Looks good to me, basically a one-to-one migration. The single binary will have to set up the signal handler on the context and starting the profiling server if desired.

kumare3:
> Will defer on the initCerts question. Looks good to me, basically a one-to-one migration. The single binary will have to set up the signal handler on the context and starting the profiling server if desired.

Exactly, the profiling server, signal handler can be independent

EngHabu (approved):

PR #400, opened 2022-02-11
Implement GetGeneratedNameWith

EngHabu:

# TL;DR
Implements TaskExecutionID.GetGeneratedNameWith 

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue


wild-endeavor (dismissed):

pkg/controller/nodes/task/taskexec_context.go
```diff
@@ -43,6 +44,19 @@ func (te taskExecutionID) GetGeneratedName() string {
 	return te.execName
 }
 
+func (te taskExecutionID) GetGeneratedNameWith(minLength, maxLength int) (string, error) {
```
hamersaw, line 47: Does it makes sense to use the GetGeneratedName function rather than te.execName in this function? Then an implementation can override the GetGeneratedName function (like I'm exploring in the maptask refactor) and this function will still work.

EngHabu: If I understand you correctly, then I don't think this will work... Go doesn't have inheritance. It rather supports extending types.. does [this example](https://go.dev/play/p/jiCZ6hqmsuC) show what you have in mind?

wild-endeavor (approved):

PR #405, opened 2022-02-25
Only require default plugin configuration for task types with multiple registered plugins

hamersaw:
# TL;DR
There was a corner case where the aws_batch and k8s-array plugins could not be registered simultaneously. This fixes it.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
With this change the registration of plugins for specific task types occurs such that:
(1) task types with only one registered plugin are automatically defaulted to that plugin
(2) task types with multiple registered plugins require the default-for-task-type configuration to define the default plugin

Default plugin information resident in the PluginEntry for each plugin is no longer used in determining the default plugin for a task type.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2205

## Follow-up issue
_NA_


pkg/controller/nodes/task/config/config.go
```diff
@@ -50,7 +50,7 @@ type BarrierConfig struct {
 }
 
 type TaskPluginConfig struct {
-	EnabledPlugins []string `json:"enabled-plugins" pflag:",deprecated"`
+	EnabledPlugins []string `json:"enabled-plugins" pflag:",Plugins enabled currently"`
```
EngHabu, line 53: I'm confused... before I dig into code, do you know why are there two "EnabledPlugins" configs? one in plugins that you just modified and one here?

pkg/controller/nodes/task/plugin_config.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/flyteorg/flytepropeller/pkg/controller/nodes/task/k8s"
 )
 
-func WranglePluginsAndGenerateFinalList(ctx context.Context, cfg *config.TaskPluginConfig, pr PluginRegistryIface) ([]core.PluginEntry, error) {
+func WranglePluginsAndGenerateFinalList(ctx context.Context, cfg *config.TaskPluginConfig, pr PluginRegistryIface) ([]core.PluginEntry, map[pluginID][]taskType, error) {
```
EngHabu, line 16: nit:
```suggestion
func WranglePluginsAndGenerateFinalList(ctx context.Context, cfg *config.TaskPluginConfig, pr PluginRegistryIface) (enabledPlugins []core.PluginEntry, defaultForTaskTypes map[pluginID][]taskType, err error) {
```

pkg/controller/nodes/task/config/config.go
hamersaw: I'm not sure exactly. Both the [plugins config](https://github.com/flyteorg/flyteplugins/blame/5a80da6b9d9a6fbca413148aaf712e5615be6e52/go/tasks/config/config.go#L20) and [propeller config](https://github.com/flyteorg/flytepropeller/pull/1) have been there since the initial commit. The propeller [default plugin for task type PR](https://github.com/flyteorg/flytepropeller/pull/199) change the EnabledPlugins comment to include "deprecated" but then still used the field - I wonder if the wrong repository was deprecated? I don't see the plugins config field used anywhere.

pkg/controller/nodes/task/plugin_config.go
hamersaw: done

pkg/controller/nodes/task/config/config.go
EngHabu: Looks like the plugins' EnabledPlugins field is not used. Maybe we should delete that then...

pkg/controller/nodes/task/plugin_config.go
```diff
@@ -13,23 +13,21 @@ import (
...
+		AllDefaultForTaskTypes: defaultForTaskTypes,
 	}
-	var err error
 	if cfg != nil {
 		pluginsConfigMeta, err = cfg.GetEnabledPlugins()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 	}
```
EngHabu, line 26: I'm trying to read through this function and I think the logic here is a bit twisted (I understand it's not part of the PR but just trying to leave the code a bit better than when we started)... 
In what scenarios do we pass a nil `cfg` and why do we expect that to run through the function as opposed to erroring...

This block feels like it should be re-written to something like:
```golang
    if cfg == nil {
        return nil, nil, fmt.Errorf("cfg is a required argument")
    }

    pluginsConfigMeta, err := cfg.GetEnabledPlugins()
    if err != nil {
        return nil, nil, err
    }

EngHabu (approved):
nits

pkg/controller/nodes/task/config/config.go
hamersaw: Sure, took care of it in the [flyteplugins pr](https://github.com/flyteorg/flyteplugins/pull/248).

pkg/controller/nodes/task/plugin_config.go
hamersaw: Since we default the TaskPlugins config to [empty lists](https://github.com/flyteorg/flytepropeller/blob/3baef11bb401c1bf7523e26d3423e356814e5e00/pkg/controller/nodes/task/config/config.go#L21) there is no place where we can get nil within FlytePropeller. I validated by passing empty configs on startup. I updated to error on a nil config.

EngHabu (approved):

PR #410, opened 2022-03-03
Fix: the pod labels in the podTemplate are ignored by the propeller manager

mayitbeegh:
# TL;DR
The flytepropeller manager ignores the pod labels in the flytepropeller's podTemplate, resulting in discrepancies in pod labels between pods created by Helm and pods deployed by flyepropeller manager.
Example labels:
Created by Helm:
```
labels:
    app.kubernetes.io/instance: foo-helmchart
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: flytepropeller
    helm.sh/chart: bar
    pod-template-hash: foobar
```
Created by flytepropeller manager:
```
labels:
    app: flytepropeller
```
This PR tries to add labels defined in podTemplate to the manager deployed pods while maintaining backwards compatibility. Example labels after the fix:
Created by flytepropeller manager
```
labels:
    app.kubernetes.io/instance: foo-helmchart
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: flytepropeller
    helm.sh/chart: bar
    pod-template-hash: foobar
    app: flytepropeller
```

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Follow-up issue
_NA_

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -86,15 +86,14 @@ func (m *Manager) createPods(ctx context.Context) error {
 		"shardConfigHash":            fmt.Sprintf("%d", shardConfigHash),
 	}
 	podNames := m.getPodNames()
-	podLabels := map[string]string{
-		"app": m.podApplication,
-	}
-	// The "app" label might be overwritten by the pod template
+	podLabels := make(map[string]string)
 	if podTemplate.Template.Labels != nil {
 		for k, v := range podTemplate.Template.Labels {
 			podLabels[k] = v
 		}
 	}
+	// The "app" label from podTemplate might be overwritten by this
+	podLabels["app"] = m.podApplication
 	// disable leader election on all managed pods
 	container, err := utils.GetContainer(&podTemplate.Template.Spec, m.podTemplateContainerName)
 	if err != nil {
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -89,7 +89,12 @@ func (m *Manager) createPods(ctx context.Context) error {
 	podLabels := map[string]string{
 		"app": m.podApplication,
 	}
-
+	// The "app" label might be overwritten by the pod template
+	if podTemplate.Template.Labels != nil {
+		for k, v := range podTemplate.Template.Labels {
+			podLabels[k] = v
+		}
+	}
 	// disable leader election on all managed pods
 	container, err := utils.GetContainer(&podTemplate.Template.Spec, m.podTemplateContainerName)
 	if err != nil {


manager/manager.go
```diff
@@ -89,7 +89,12 @@ func (m *Manager) createPods(ctx context.Context) error {
 	podLabels := map[string]string{
 		"app": m.podApplication,
 	}
-
+	// The "app" label might be overwritten by the pod template
+	if podTemplate.Template.Labels != nil {
+		for k, v := range podTemplate.Template.Labels {
+			podLabels[k] = v
+		}
```
EngHabu, line 96: I think we should set the `"app": m.podApplication,` after this... manager relies on the app label to find these pods... we shouldn't allow bad podTemplates to cause infinite pod creation

EngHabu (commented):
@hamersaw please review

manager/manager.go
mayitbeegh: Fixed

manager/manager.go
```diff
@@ -89,7 +89,12 @@ func (m *Manager) createPods(ctx context.Context) error {
 	podLabels := map[string]string{
 		"app": m.podApplication,
 	}
-
+	// The "app" label might be overwritten by the pod template
+	if podTemplate.Template.Labels != nil {
```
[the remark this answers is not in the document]
mayitbeegh: I think we should unless there's a reason to deviate from the podTemplate. The only annotation from podTemplate is `configChecksum` for now.

hamersaw:
This is a great find! I think there is an easy fix.

In the new default PodTemplate PR we're copying the [ObjectMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) from the PodTemplate.Template to the actual Pod [here](https://github.com/flyteorg/flyteplugins/blob/de97292a05ac9e4e4902a0a2380e66877b728d52/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L167). This approach ensures that everything defined in the PodTemplate is correctly transfered, including labels and annotations. Right now the manager tracks labels and annotations separately, I think we should have a variable, like `objectMeta` that is a deep copy of the PodTemplate.ObjectMeta instead of this, then override the necessary values (ex. specific labels & annotations, pod name, pod namespace, etc defined [here](https://github.com/flyteorg/flytepropeller/blob/048c8653f2086f0bcfc61941e0814c23418f2af5/manager/manager.go#L172-L176). This way we ensure that the managed FlytePropeller pods follow the PodTemplate as closely as possible.

mayitbeegh:
> This is a great find! I think there is an easy fix.
> 
> In the new default PodTemplate PR we're copying the [ObjectMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) from the PodTemplate.Template to the actual Pod [here](https://github.com/flyteorg/flyteplugins/blob/de97292a05ac9e4e4902a0a2380e66877b728d52/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L167). This approach ensures that everything defined in the PodTemplate is correctly transfered, including labels and annotations. Right now the manager tracks labels and annotations separately, I think we should have a variable, like `objectMeta` that is a deep copy of the PodTemplate.ObjectMeta instead of this, then override the necessary values (ex. specific labels & annotations, pod name, pod namespace, etc defined [here](https://github.com/flyteorg/flytepropeller/blob/048c8653f2086f0bcfc61941e0814c23418f2af5/manager/manager.go#L172-L176). This way we ensure that the managed FlytePropeller pods follow the PodTemplate as closely as possible.

Should we close this PR then in favor of the PodTemplate PR? @EngHabu @hamersaw

hamersaw:
> > This is a great find! I think there is an easy fix.
> > In the new default PodTemplate PR we're copying the [ObjectMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) from the PodTemplate.Template to the actual Pod [here](https://github.com/flyteorg/flyteplugins/blob/de97292a05ac9e4e4902a0a2380e66877b728d52/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L167). This approach ensures that everything defined in the PodTemplate is correctly transfered, including labels and annotations. Right now the manager tracks labels and annotations separately, I think we should have a variable, like `objectMeta` that is a deep copy of the PodTemplate.ObjectMeta instead of this, then override the necessary values (ex. specific labels & annotations, pod name, pod namespace, etc defined [here](https://github.com/flyteorg/flytepropeller/blob/048c8653f2086f0bcfc61941e0814c23418f2af5/manager/manager.go#L172-L176). This way we ensure that the managed FlytePropeller pods follow the PodTemplate as closely as possible.
> 
> Should we close this PR then in favor of the PodTemplate PR? @EngHabu @hamersaw

Oh no, sorry. The default PodTemplate work is solving a different problem. I think that in working with PodTemplates more it might be a better, translatable way to handle labels, annotations, etc. In hindsight it would have been a better solution to handle the PodTemplate here in the FlytePropeller manager that way from the start.

mayitbeegh:
> > > This is a great find! I think there is an easy fix.
> > > In the new default PodTemplate PR we're copying the [ObjectMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) from the PodTemplate.Template to the actual Pod [here](https://github.com/flyteorg/flyteplugins/blob/de97292a05ac9e4e4902a0a2380e66877b728d52/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L167). This approach ensures that everything defined in the PodTemplate is correctly transfered, including labels and annotations. Right now the manager tracks labels and annotations separately, I think we should have a variable, like `objectMeta` that is a deep copy of the PodTemplate.ObjectMeta instead of this, then override the necessary values (ex. specific labels & annotations, pod name, pod namespace, etc defined [here](https://github.com/flyteorg/flytepropeller/blob/048c8653f2086f0bcfc61941e0814c23418f2af5/manager/manager.go#L172-L176). This way we ensure that the managed FlytePropeller pods follow the PodTemplate as closely as possible.
> > 
> > 
> > Should we close this PR then in favor of the PodTemplate PR? @EngHabu @hamersaw
> 
> Oh no, sorry. The default PodTemplate work is solving a different problem. I think that in working with PodTemplates more it might be a better, translatable way to handle labels, annotations, etc. In hindsight it would have been a better solution to handle the PodTemplate here in the FlytePropeller manager that way from the start.

If I understand it correctly, there's another way of handling PodTemplate by copying everything to an `ObjectMeta` field, with you linked, that works better than how flytepropeller-manager handles annotations, labels, etc separately. Is that correct? If so, perhaps we can open up an issue to refactor flytepropeller-manager in the future, and proceed with this PR as a temporary Band-aid™.

mayitbeegh:
Closed in favor of https://github.com/flyteorg/flytepropeller/pull/411

PR #412, opened 2022-03-09
Add Architecture field to support ARM instances

reverson:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_


reverson:
@akhurana001

pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -191,6 +191,11 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 		interruptible = wf.GetMetadataDefaults().GetInterruptible()
 	}
 
+	var architecture core.Container_Architecture
+	for _, t := range tasks {
```
akhurana001, line 195: this looks a bit weird. We just pick up the flag for the last task ?

reverson: I agree that's weird.  I can update it to set the architecture only if it's set in one of the tasks and then break, which would cause all the tasks to run on the same architecture, but that would be necessary anyway, right? Tasks don't run on their own nodes, they all run within the same node, correct?

PR #419, opened 2022-03-30
Fix the Garbage collection problem

aeioulisa:
# TL;DR
The default setting of 23 hours for ttl in flytepropeller, seems to leave certain workflows behind.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

The way we used to identify which wf should be thrown is:
1. mark the complete hour of the day when the wf complete
2. list the hours to be deleted when the GC is running, this list is calculated based on ttl and the hour of the current time:

```
 	list = [0,1,2,...,23]
	list.remove(now ~ now-ttl)
```
The problem is that the more ttl larger, the shorter list is. For example, if ttl=23 and now is 13:00, the list will be [14]; ttl=23 and now is 1:00, the list will be [2]. That means the GC will only delete the wf whose complete time is 14 or 2 forever.
Besides, we only have 24 indexes, but the wf may live more than 24h as we accept the case of  ttl+gcinterval>24. 

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1554

## Follow-up issue
_NA_


pkg/controller/completed_workflows.go
```diff
@@ -56,32 +56,24 @@ func HasCompletedLabel(w *v1alpha1.FlyteWorkflow) bool {
 
 // Calculates a list of all the hours that should be deleted given the current hour of the day and the retentionperiod in hours
 // Usually this is a list of all hours out of the 24 hours in the day - retention period - the current hour of the day
-func CalculateHoursToDelete(retentionPeriodHours, currentHourOfDay int) []string {
```
pingsutw, line 59: We should keep this function for backward compatibility

pkg/controller/completed_workflows.go
```diff
@@ -11,7 +11,7 @@ import (
 const controllerAgentName = "flyteworkflow-controller"
 const workflowTerminationStatusKey = "termination-status"
 const workflowTerminatedValue = "terminated"
-const hourOfDayCompletedKey = "hour-of-day"
+const CompletedTimeKey = "hour-of-day"
```
pingsutw, line 14: could we revert it?

pkg/controller/garbage_collector_test.go
```diff
@@ -101,7 +101,7 @@ func TestGarbageCollector_StartGC(t *testing.T) {
 		DeleteCollectionCb: func(options *v1.DeleteOptions, listOptions v1.ListOptions) error {
 			assert.NotNil(t, options)
 			assert.NotNil(t, listOptions)
-			assert.Equal(t, "hour-of-day in (0,1,10,11,12,13,14,15,16,17,18,19,2,20,21,3,4,5,6,7,8,9),termination-status=terminated", listOptions.LabelSelector)
+			assert.Equal(t, "hour-of-day notin (10.22,10.23),termination-status=terminated", listOptions.LabelSelector)
```
pingsutw, line 104:
```suggestion
			assert.Equal(t, "hour-of-day not in (10.22,10.23),termination-status=terminated", listOptions.LabelSelector)
```

aeioulisa: Hi @pingsutw, thanks for your review. 
according to [k8s doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/):
Set-based label requirements allow filtering keys according to a set of values. Three kinds of operators are supported: in,notin and exists

pkg/controller/completed_workflows.go
```diff
@@ -73,15 +73,25 @@ func CalculateHoursToDelete(retentionPeriodHours, currentHourOfDay int) []string
...
 func CompletedWorkflowsSelectorOutsideRetentionPeriod(retentionPeriodHours int, currentTime time.Time) *v1.LabelSelector {
-	hoursToDelete := CalculateHoursToDelete(retentionPeriodHours, currentTime.Hour())
+	hoursToKeep := CalculateHoursToKeep(retentionPeriodHours, currentTime)
 	s := CompletedWorkflowsLabelSelector()
 	s.MatchExpressions = append(s.MatchExpressions, v1.LabelSelectorRequirement{
 		Key:      hourOfDayCompletedKey,
-		Operator: v1.LabelSelectorOpIn,
-		Values:   hoursToDelete,
+		Operator: v1.LabelSelectorOpNotIn,
+		Values:   hoursToKeep,
```
EngHabu, line 94: I wrote up a comment that got lost...

I've a question about migration. What happens when this version of propeller is installed on an existing cluster? It seems to me that it'll start deleting all completed workflows regardless of the "HoursToKeep" because non of them will match the new labels... Can you think of a way to not break that?

aeioulisa: hi @EngHabu, thanks for your review.
Maybe we can keep the old label selector temporary to not break that. But in this case, the workflows that are supposed to leave but have not will still be stuck

pkg/controller/completed_workflows.go
```diff
@@ -73,15 +75,26 @@ func CalculateHoursToDelete(retentionPeriodHours, currentHourOfDay int) []string
 	return hoursToDelete
 }
 
+//Calculates a list of all the hours that should be kept given the current time and the retentionperiod in hours
+func CalculateHoursToKeep(retentionPeriodHours int, currentTime time.Time) []string {
+	hoursToKeep := make([]string, 0, retentionPeriodHours+1)
+	for i := 0; i <= retentionPeriodHours; i++ {
+		hoursToKeep = append(hoursToKeep, strings.ReplaceAll(
+			strings.Split(currentTime.Round(time.Hour).String(), ":")[0], " ", "."))
```
EngHabu, line 83: Can you elaborate on why are we splitting by ":" here and by " +" in line 46?

pkg/controller/completed_workflows.go
aeioulisa: done

pkg/controller/completed_workflows.go
EngHabu: Not: maybe you can move that logic (the split and replace) to a simple function and use that in both places to make sure the behavior is always consistent

EngHabu (commented):
A couple of nits

pkg/controller/config/config.go
```diff
@@ -53,9 +53,9 @@ var (
 			Duration: 30 * time.Second,
 		},
 		MaxWorkflowRetries: 10,
-		MaxTTLInHours:      23,
+		MaxTTLInHours:      1,
```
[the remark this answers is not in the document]
aeioulisa: Should we keep the GCInterval 30 minutes or set a longer time?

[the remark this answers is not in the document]
EngHabu: 30 mins is fine...

pkg/controller/completed_workflows.go
EngHabu: How did you deal with the case you described? The old stuck workflows?

aeioulisa: The new commits only solve the problem of deleting all completed workflows when this propeller version is installed. The old stuck workflows can be deleted by reducing GCInterval. It will not be a problem if GCInterval is 30 mins.

EngHabu (approved):
Thank you, Lisa!

pkg/controller/completed_workflows.go
```diff
@@ -73,10 +75,36 @@ func CalculateHoursToDelete(retentionPeriodHours, currentHourOfDay int) []string
...
 func CompletedWorkflowsSelectorOutsideRetentionPeriod(retentionPeriodHours int, currentTime time.Time) *v1.LabelSelector {
-	hoursToDelete := CalculateHoursToDelete(retentionPeriodHours, currentTime.Hour())
+	hoursToKeep := CalculateHoursToKeep(retentionPeriodHours, currentTime)
+	s := CompletedWorkflowsLabelSelector()
+	s.MatchExpressions = append(s.MatchExpressions, v1.LabelSelectorRequirement{
+		Key:      completedTimeKey,
+		Operator: v1.LabelSelectorOpNotIn,
+		Values:   hoursToKeep,
+	})
+	s.MatchExpressions = append(s.MatchExpressions, v1.LabelSelectorRequirement{
```
honnix, line 98: It seems this additional condition would result not matching anything because new resources have no `hour-of-day` label.

PR #429, opened 2022-04-21
Apply interruptible override from workflow execution config #minor

MorpheusXAUT:
# TL;DR
Applies `interruptible` override from workflow execution spec to nodes, allowing for workflows to be flagged as interruptible for a single execution.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
The `ExecutionConfig` of a `FlyteWorkflow` now contains an `Interruptible` flag, indicating a workflow should be marked as interruptible for a single execution. The `IsInterruptible` getter of `FlyteWorkflow` has been adapted to return either the new `Interruptible` flag of the `ExecutionConfig` or the already existing one from the `NodeDefaults`, maintaining backward compatibility. Furthermore, the existing defaults logic for interruptible tasks still applies - a task defined with `interruptible=False` will still remain non-schedulable for spot instances.
Two minor "interruptible" typos in the `NodeSpec` and `nodeExecMetadata` types have been fixed - changes are backward compatible since the json representation had already been spelled correctly.

Tests have been added to verify correct application of interruptible "inheritance" for nodes, smoke testing has been performed manually for workflow/single task executions and relaunches together with the flyteadmin changes.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2284

## Follow-up issue
_NA_


MorpheusXAUT:
Whilst a workflow can now have its interruptible setting overwritten to `false`, the task settings still take preference - if a task is defined with `interruptible=True`, it will _not_ be overwritten to `false`, even if the workflow execution config has `Interruptible: false` specified.
If perform a single execution with an override disabling spot-instances should also overrule all task preferences, some further changes are needed.

EngHabu:
@MorpheusXAUT thank you for your contribution!

@hamersaw, mind taking a look at this?

hamersaw (commented):
If I understand correctly, the workflow interruptible flag only applies as a default in the case that interruptible is not explicitly set on the node. Looks like this functionality is [quite old](https://github.com/flyteorg/flytepropeller/commit/38ba9c10ccadc2e9644d9a07f7602f82ab92d8bd). I don't think this is necessarily intuitive in some cases. For example, setting the workflow interruptible to true means all nodes are interruptible (including those already set) but if a user sets the workflow interruptible to false some nodes may still be interrupted because it's only applied as a default.

Not saying this is a change we should make, but if we are further exposing this functionality (in the UI changes) perhaps it warrants a brief discussion. Have you thought about this? My first thought is it may make more sense to handle as follows:
workflow interruptible = true then all nodes are interruptible
workflow interruptible not set then use the interruptible set on nodes
workflow interruptible = false then all nodes are non-interruptible

I believe it would only require a simple change to [this conditional](https://github.com/flyteorg/flytepropeller/blob/24c22006721ebd8d389d242f997001cec586b9bc/pkg/controller/nodes/node_exec_context.go#L212-L215).

MorpheusXAUT:
> If I understand correctly, the workflow interruptible flag only applies as a default in the case that interruptible is not explicitly set on the node. Looks like this functionality is [quite old](https://github.com/flyteorg/flytepropeller/commit/38ba9c10ccadc2e9644d9a07f7602f82ab92d8bd). I don't think this is necessarily intuitive in some cases. For example, if the workflow interruptible is set to false some nodes may still be set as interruptible.

That's correct - the interruptible flag on the node currently has the "last call" when it comes to picking a setting.

> Not saying this is a change we should make, but if we are further exposing this functionality (in the UI changes) perhaps it warrants a brief discussion. Have you thought about this? My first thought is it may make more sense to handle as follows:
workflow interruptible = true then all nodes are interruptible
workflow interruptible not set then use the interruptible set on nodes
workflow interruptible = false then all nodes are non-interruptible

I did think about that and initially had the override "skip" the node settings just like you suggested, but actually chose to keep the current behavior after talking to our team about what their expectations would be.
Our main concern point was with tasks that cannot properly be run as interruptible (e.g. because of side effects) being overriden because the person running them might not fully know about these pitfalls. Allowing the person writing the workflow to explicitly disabling the interruptible setting _no matter what_ lets them ensure there's no issues in the future.

For that reason, I wouldn't force every node to run as interruptible when using the workflow override, although I agree that it might be a bit unintuitive to users in some case.
The other way around - forcing some nodes that "want" to be interruptible to not be - would not be as "problematic" I'd say, but the question is if adding just that wouldn't be even more inconsistent...

> I believe it would only require a simple change to [this conditional](https://github.com/flyteorg/flytepropeller/blob/24c22006721ebd8d389d242f997001cec586b9bc/pkg/controller/nodes/node_exec_context.go#L212-L215).

Yeah, that should do it, although we'd probably also need to change the signature of [`IsInterruptible()`](https://github.com/flyteorg/flytepropeller/blob/24c22006721ebd8d389d242f997001cec586b9bc/pkg/apis/flyteworkflow/v1alpha1/iface.go#L444) to return a `*bool` so we can distinguish between no override being provided and it being set to `false` - not sure how many other changes that would require off the top of my head.

hamersaw (approved):

PR #430, opened 2022-04-22
GC: Fix condition for finding completed workflows #major

EngHabu:

# TL;DR
The condition for finding completed (terminated) workflows that are ready to be garbage collected is over constrained and therefore never finds any workflows. 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2400
https://github.com/flyteorg/flyte/issues/2396

### DIFF
--- a/pkg/controller/completed_workflows.go
+++ b/pkg/controller/completed_workflows.go
@@ -16,7 +16,7 @@ const (
 	completedTimeKey             = "completed-time"
 	// Layout string for time.Format() function expects the time ("Mon, 02 Jan 2006 15:04:05 MST") formatted
 	// in the desired layout.
-	labelTimeFormat = "2006-02-01.15"
+	labelTimeFormat = "2006-01-02.15"
 )
 
 // IgnoreCompletedWorkflowsLabelSelector this function creates a label selector, that will ignore all objects (in this
--- a/pkg/controller/completed_workflows.go
+++ b/pkg/controller/completed_workflows.go
@@ -2,18 +2,22 @@ package controller
 
 import (
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const controllerAgentName = "flyteworkflow-controller"
-const workflowTerminationStatusKey = "termination-status"
-const workflowTerminatedValue = "terminated"
-const hourOfDayCompletedKey = "hour-of-day"
-const completedTimeKey = "completed-time"
+const (
+	controllerAgentName          = "flyteworkflow-controller"
+	workflowTerminationStatusKey = "termination-status"
+	workflowTerminatedValue      = "terminated"
+	hourOfDayCompletedKey        = "hour-of-day"
+	completedTimeKey             = "completed-time"
+	// Layout string for time.Format() function expects the time ("Mon, 02 Jan 2006 15:04:05 MST") formatted
+	// in the desired layout.
+	labelTimeFormat = "2006-02-01.15"
+)
 
 // IgnoreCompletedWorkflowsLabelSelector this function creates a label selector, that will ignore all objects (in this
 // case workflow) that DOES NOT have a label key=workflowTerminationStatusKey with a value=workflowTerminatedValue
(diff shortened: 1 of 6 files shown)


honnix (dismissed):

pkg/controller/completed_workflows.go
```diff
@@ -95,26 +97,25 @@ func CompletedWorkflowsSelectorOutsideRetentionPeriod(retentionPeriodHours int,
 		Operator: v1.LabelSelectorOpNotIn,
 		Values:   hoursToKeep,
 	})
-	s.MatchExpressions = append(s.MatchExpressions, v1.LabelSelectorRequirement{
-		Key:      hourOfDayCompletedKey,
-		Operator: v1.LabelSelectorOpIn,
-		Values:   []string{""},
-	})
```
aeioulisa, line 102: sorry for this mistake 
I think all completed workflows will be contained if we remove this match expression.
change operator to v1.LabelSelectorOpDoesNotExist can solve this problem

EngHabu: All "old" completed workflows, right? should that be ok given we are trying to fix this existing bug?

pkg/controller/completed_workflows.go
```diff
@@ -95,26 +97,25 @@ func CompletedWorkflowsSelectorOutsideRetentionPeriod(retentionPeriodHours int,
...
 		Values:   hoursToDelete,
 	})
+
 	return s
 }
 
-func SplitTimeToMeetsLabelFormat(currentTime time.Time) string {
+func FormatTimeForLabel(currentTime time.Time) string {
 	return strings.ReplaceAll(
 		strings.Split(currentTime.Round(time.Hour).String(), ":")[0], " ", ".")
```
hamersaw, line 120: Can we make this currentTime.Hour()? It should work fine as is because the same function is used to set the completed label time, but there's something about a completed time of 11:31 being labeled as the hour 12 that doesn't seem right.

pkg/controller/completed_workflows.go
hamersaw: I think this addresses [this comment](https://github.com/flyteorg/flytepropeller/pull/419/files#r839975491) right? It seems to me that the v1.LabelSelectorOpDoesNotExist solution that @aeioulisa suggested would work. Basically, since we are calling both [the old](https://github.com/flyteorg/flytepropeller/blob/4e4df8a4e520067b2461a42393ddae71e82c52e6/pkg/controller/garbage_collector.go#L136) and [the new](https://github.com/flyteorg/flytepropeller/blob/4e4df8a4e520067b2461a42393ddae71e82c52e6/pkg/controller/garbage_collector.go#L140) garbage collection functions on each iteration, this would make sure that each only processes workflow CRDs that it is responsible for (ie. old function for old labels and new function for new labels).

honnix: I agree that we should add `v1.LabelSelectorOpDoesNotExist` to avoid garbage collecting things in the wrong iteration.

EngHabu: My understanding is that the issue with the existing delete is that it misses out some workflows indefinitely... this would require a one-time delete operation for these workflows... are you all saying this should just go in documentation, @hamersaw, @honnix, @aeioulisa ?

hamersaw: Oh OK, I may not have the context for old workflow issues. I don't think it is fair to expect users to manually cleanup old workflows. So this solution will cleanup all old workflows and only recognize the retention period when GCing workflow CRDs moving forward?

pkg/controller/completed_workflows.go
EngHabu: `currentTime.Hour()` returns an int for the hour within the day... so if you have a retention of 25 hours or something, you wouldn't know whether a label of value "1" has completed today or yesterday... no?

> but there's something about a completed time of 11:31 being labeled as the hour 12 that doesn't seem right.
Similarly, 11:59 will be in the same hour as 11:01.. I think we should just document that to say "deletion will happen +/- an hour of the retention period". 
We can make this more granular (per half hour...) I don't know what's the limit (size of request, perf... etc.) when it comes to the matching expression...

hamersaw: Sure, I figured it would be used in on conjunction with the Year(), Month(), and Day() functions to output identical to the current function (ie. "yyyy-mm-dd.hh"). Very nit, OK with not making the change.

hamersaw (approved):

pkg/controller/completed_workflows.go
honnix: I feel it might be better we document the manual step to clean up, instead of deleting all old workflows without respecting `hour-of-day`.

Another thing. If we do not check existence of `hour-of-day` together with checking `completed-time`, we might just remove this iteration: https://github.com/flyteorg/flytepropeller/blob/4e4df8a4e520067b2461a42393ddae71e82c52e6/pkg/controller/garbage_collector.go#L136 because everything not having `completed-time` or out of retention period will be cleaned up.

EngHabu: @hamersaw, yeah it's not my first choice either... will at least include that in the release notes...
@honnix updated with the additional filter...

honnix: Nice, thanks!

honnix: I think the test case needs to be updated.

pkg/controller/completed_workflows.go
EngHabu: Agreed... addressed... better?

hamersaw (approved):

PR #433, opened 2022-04-28
Empty lists/maps should match list/maps of any types

EngHabu:

# TL;DR
Treat empty lists/maps differently from how Void/None is treated. Empty lists/maps should be assignable to any lists/maps where as Void should only be assignable to NoneType...

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2424

### DIFF
--- a/pkg/compiler/validators/typing.go
+++ b/pkg/compiler/validators/typing.go
@@ -61,12 +61,9 @@ type mapTypeChecker struct {
 // and value types match.
 func (t mapTypeChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
 	// Empty maps should match any collection.
-	if isNoneType(upstreamType.GetMapValueType()) {
+	if mapLiteralType := upstreamType.GetMapValueType(); isNoneType(mapLiteralType) {
 		return true
-	}
-
-	mapLiteralType := upstreamType.GetMapValueType()
-	if mapLiteralType != nil {
+	} else if mapLiteralType != nil {
 		return getTypeChecker(t.literalType.GetMapValueType()).CastsFrom(mapLiteralType)
 	}
 
--- a/pkg/compiler/validators/typing.go
+++ b/pkg/compiler/validators/typing.go
@@ -79,14 +79,13 @@ type collectionTypeChecker struct {
 // subtypes match.
 func (t collectionTypeChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
 	// Empty collections should match any collection.
+	collectionType := upstreamType.GetCollectionType()
 	if isNoneType(upstreamType.GetCollectionType()) {
 		return true
-	}
-
-	collectionType := upstreamType.GetCollectionType()
-	if collectionType != nil {
+	} else if collectionType != nil {
 		return getTypeChecker(t.literalType.GetCollectionType()).CastsFrom(collectionType)
 	}
+
 	return false
 }
 
--- a/pkg/compiler/validators/typing.go
+++ b/pkg/compiler/validators/typing.go
@@ -15,29 +15,7 @@ type trivialChecker struct {
 	literalType *flyte.LiteralType
 }
 
-type voidChecker struct{}
-
-type mapTypeChecker struct {
-	literalType *flyte.LiteralType
-}
-
-type collectionTypeChecker struct {
-	literalType *flyte.LiteralType
-}
-
-type schemaTypeChecker struct {
-	literalType *flyte.LiteralType
-}
-
-type unionTypeChecker struct {
-	literalType *flyte.LiteralType
-}
-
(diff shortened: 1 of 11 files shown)


pkg/compiler/validators/utils.go
```diff
@@ -153,43 +153,72 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
+
 			if innerType != nil && !AreTypesCastable(otherType, innerType) {
-				return &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_NONE}}
+				return &core.LiteralType{
+					Type: &core.LiteralType_CollectionType{
+						CollectionType: &core.LiteralType{
+							Type: &core.LiteralType_Simple{Simple: core.SimpleType_NONE},
+						},
+					},
+				}
```
hamersaw, line 185: I'm not sure I fully understand this. In this case if we find that the collection is not homogenous (ex. [9.99, "foo"]) then return [None], but this change explicitly allows for [None] to be castable to any type (ex. [Int], [String], etc). Is there a potential bug here where [9.99, "foo"] is castable to [Int]? Can we add a unit test for this case?

pkg/compiler/validators/utils.go
```diff
@@ -153,43 +153,72 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
 			otherType := LiteralTypeForLiteral(x)
 			if innerType != nil && !AreTypesCastable(otherType, innerType) {
-				return &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_NONE}}
+				return &core.LiteralType{
+					Type: &core.LiteralType_MapValueType{
+						MapValueType: &core.LiteralType{
+							Type: &core.LiteralType_Simple{Simple: core.SimpleType_NONE},
+						},
+					},
+				}
```
hamersaw, line 216: Same comment as a non-homogenous list being castable to any list above.

pkg/compiler/validators/typing.go
```diff
@@ -68,29 +46,57 @@ func (t trivialChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
...
+	if isNoneType(upstreamType.GetMapValueType()) {
+		return true
+	}
+
 	mapLiteralType := upstreamType.GetMapValueType()
 	if mapLiteralType != nil {
 		return getTypeChecker(t.literalType.GetMapValueType()).CastsFrom(mapLiteralType)
 	}
+
 	return false
```
hamersaw, line 73: Does this make it a little more readable?
```suggestion
	mapLiteralType := upstreamType.GetMapValueType()
	if isNoneType(mapLiteralType) {
		// Empty maps should match any collection.
		return true
	} else if mapLiteralType != nil {
		return getTypeChecker(t.literalType.GetMapValueType()).CastsFrom(mapLiteralType)
	}

	return false
```

pkg/compiler/validators/typing.go
```diff
@@ -68,29 +46,57 @@ func (t trivialChecker) CastsFrom(upstreamType *flyte.LiteralType) bool {
...
+	// Empty collections should match any collection.
+	if isNoneType(upstreamType.GetCollectionType()) {
+		return true
+	}
+
 	collectionType := upstreamType.GetCollectionType()
 	if collectionType != nil {
 		return getTypeChecker(t.literalType.GetCollectionType()).CastsFrom(collectionType)
 	}
 	return false
```
hamersaw, line 92: If we decide to reformat the mapTypeChecker.CastsFrom about then this block should probably mimic the style.

pkg/compiler/validators/utils.go
EngHabu: so originally I left it to that yes because Array types support this, very awkward for various reasons, feature where you can specify minimum successes (or success ratio). When some tasks fail in an array task, their results will be substituted with `None` in the final assembled array output. This means a consuming task may receive `[10, None, 1, 3, 4]` even when it doesn't say `List[Optional[int]]` as the type... so I thought I would leave that behavior in..

However, now that you asked the question, I went back and re-read through and we don't actually go through this path for array tasks... so this is ok... 

I updated the code to, I think more correctly, build a `List[Union[int, string]]` in case the input is `[10, "foo"]` or `List[Union[int, None]]` (essentially `List[Optional[int]]` in case of `[10, None]`... wdyt?

hamersaw: Perfect, I think this makes more sense!

hamersaw (approved):

PR #436, opened 2022-05-04
Add support for GateNode with signal and sleep condition

hamersaw:
# TL;DR
Adding support for GateNodes. These nodes block execution of the downstream based on a condition. Initially, we will implement sleeping, which block for a predefined duration, and manual signals, which require a user to notify that execution may proceed. Signals may also carry values, so a user can pass a variable which may be used in downstream nodes.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/208

## Follow-up issue
_NA_


hamersaw:
cc @kumare3 @EngHabu the only thing I am not sure on here is configuration of the output variable name from a signal. All the binding lookups are performed using input variable names into the actual written protobuf results. In these cases it seems like flytekit may use the python variable name? Does it make sense to set the Signal variable name in the proto?

wild-endeavor (approved):

PR #439, opened 2022-05-16
Graceful failure on ExecutionNotFound error

hamersaw:
# TL;DR
Updating the workflow abort pipeline to drop ExecutionNotFound errors. Previously, this solution could handle workflows that have not yet started, but for those that are already executing the events that are sent as part of aborting the workflow triggered a cycle.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2275

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/handler.go
+++ b/pkg/controller/handler.go
@@ -230,8 +230,9 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 
 		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not
 		// a valid state unless we are experiencing a race condition where the workflow has not yet
-		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
-		if err != nil && eventsErr.IsNotFound(err) && w.GetExecutionStatus().GetPhase() != v1alpha1.WorkflowPhaseReady {
+		// been inserted into the db. Therefore, we allow retries until the maximum number of
+		// failed attempts is reached.
+		if err != nil && eventsErr.IsNotFound(err) && w.Status.FailedAttempts > uint32(p.cfg.MaxWorkflowRetries) {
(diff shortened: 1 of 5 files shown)


pkg/controller/handler.go
```diff
@@ -230,8 +230,9 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 
 		// ExecutionNotFound error is returned when flyteadmin is missing the workflow. This is not
 		// a valid state unless we are experiencing a race condition where the workflow has not yet
-		// been inserted into the db (ie. workflow phase is WorkflowPhaseReady).
-		if err != nil && eventsErr.IsNotFound(err) && w.GetExecutionStatus().GetPhase() != v1alpha1.WorkflowPhaseReady {
+		// been inserted into the db. Therefore, we allow retries until the maximum number of
+		// failed attempts is reached.
+		if err != nil && eventsErr.IsNotFound(err) && w.Status.FailedAttempts > uint32(p.cfg.MaxWorkflowRetries) {
```
EngHabu, line 235: Won't [this](https://github.com/flyteorg/flytepropeller/blob/c22158126a1785165ab97d1ab83eaf063e244b66/pkg/controller/handler.go#L95) eventually break out anyway?

hamersaw: Yes, you are correct. This is not the right fix for this, I have done some extensive testing now and found the issue was a cyclic ExecutionNotFound error when attempted to abort the workflow. I will leave this as is.

EngHabu (approved):

PR #440, opened 2022-05-17
Add flag to enable latency metrics

sonjaer:
# TL;DR
We would like latency metrics on the outgoing RPCs. I found in the doc that they can be enabled by doing [this](https://github.com/grpc-ecosystem/go-grpc-prometheus#histograms). Histogram metrics / metrics with high cardinality can be expensive on Prometheus servers, so I added it behind a flag (see [here](https://github.com/grpc-ecosystem/go-grpc-prometheus/blob/master/client.go#L32)).

https://github.com/flyteorg/flyte/issues/2489

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/golang/protobuf v1.4.3
 	github.com/google/uuid v1.2.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.1.0
+	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/imdario/mergo v0.3.11
 	github.com/magiconair/properties v1.8.4
 	github.com/mitchellh/mapstructure v1.4.1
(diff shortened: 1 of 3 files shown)


katrogan (dismissed):

hamersaw:
Quick question on this. Is the goal to use the histograms for SLO's or just generalized observation? The [Prometheus best practices on histograms](https://prometheus.io/docs/practices/histograms/) is a great read, it has helped me understand this a bit better, including the performance implications. My brief takeaways were: if you want an SLO on quantile values use summaries (rather than histograms) and if you want to use histograms it is very important to properly configure the bucket sizes, using default values may result in all observations in a single bucket and increasing the observation cardinality (defining > 10 buckets) has significant impact on performance.

sonjaer:
> generalized observation



> Quick question on this. Is the goal to use the histograms for SLO's or just generalized observation? The [Prometheus best practices on histograms](https://prometheus.io/docs/practices/histograms/) is a great read, it has helped me understand this a bit better, including the performance implications. My brief takeaways were: if you want an SLO on quantile values use summaries (rather than histograms) and if you want to use histograms it is very important to properly configure the bucket sizes, using default values may result in all observations in a single bucket and increasing the observation cardinality (defining > 10 buckets) has significant impact on performance.

It is for generalized observation

hamersaw:
merged in [this PR](https://github.com/flyteorg/flytepropeller/pull/460)

PR #442, opened 2022-05-20
Strip Type Metadata

EngHabu:

# TL;DR
* Remove Metadata not needed at runtime from Types/Variables (structure, annotations, description). This reduces the CRD size without compromising correctness.
* Also added flytesnacks core examples there to be compiled to help with regression testing.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2516


EngHabu:
Any issue with that? Regression testing?
I realized E2E tests do not test compiler changes because they require updating and rebuilding admin... until they do, I would like some validation beforehand...

hamersaw (dismissed):
If I understand this correctly, can we remove [this section](https://github.com/flyteorg/flytepropeller/blob/24e93cd9e7930bac95968f46c0d6b6fc46bad3e3/pkg/compiler/transformers/k8s/node.go#L224-L244) in the [buildTasks](https://github.com/flyteorg/flytepropeller/blob/24e93cd9e7930bac95968f46c0d6b6fc46bad3e3/pkg/compiler/transformers/k8s/node.go#L213) function which strips annotations from the LiteralType since this PR removes them as part of the StripTypeMetadata function?

Once that is figured out, looks great! I wonder how much space this saves on the CRD.

EngHabu:
Excellent point @hamersaw

hamersaw (approved):

PR #443, opened 2022-05-24
Add deck_uri to NodeExecutionEvent

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
Add deck_uri in NodeExecutionEvent

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2175

## Follow-up issue
_NA_


pingsutw:
cc @kumare3 @EngHabu

events/node_event_recorder.go
```diff
@@ -49,6 +54,16 @@ func (r *nodeEventRecorder) handleFailure(ctx context.Context, ev *event.NodeExe
...
 	var rawOutputPolicy = eventConfig.RawOutputPolicy
+
+	if len(ev.GetOutputUri()) > 0 {
+		// Both outputs.pb and deck.html should be in the same folder
+		deckURI := strings.Replace(ev.GetOutputUri(), outputsFile, deckFile, 1)
+		metadata, err := r.store.Head(ctx, storage.DataReference(deckURI))
+		if err == nil && metadata.Exists() {
+			ev.DeckUri = deckURI
+		}
+	}
```
EngHabu, line 65: It feels wrong to do this here... I think it should happen here: pkg/controller/nodes/task/pre_post_execution.go:118.

We should then probably do it again here: https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/nodes/task/handler.go#L525 to populate the file just like we do with output.pb...

You will need to carry this information around so that when we create the event, we don't need to check for the existence of the file again...

go.mod
```diff
@@ -146,3 +145,5 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
+
+replace github.com/flyteorg/flyteplugins => github.com/flyteorg/flyteplugins v1.0.3-0.20220606160608-f19ef782f97c
```
EngHabu: Just leaving a note as a reminder to upgrade to the released version

pkg/controller/nodes/task/catalog/datacatalog/datacatalog.go
```diff
@@ -240,6 +242,21 @@ func (m *CatalogClient) Put(ctx context.Context, key catalog.Key, reader io.Outp
...
+				Scalar: &core.Scalar{
+					Value: &core.Scalar_Blob{Blob: &core.Blob{Uri: reader.GetDeckPath().String()}},
+				},
+			},
+		}
+		if outputs.Literals == nil {
+			outputs.Literals = map[string]*core.Literal{FlyteDeckFile: lt}
+		} else {
+			outputs.Literals[FlyteDeckFile] = lt
+		}
```
EngHabu, line 257: I don't feel great about just throwing it as another output... 
How about we add another level here?
like, old:
```json
{
   "o1": ...,
   "o2": ...,
   "FLYTE_DECK_FILE": ...
}
```

I'm suggesting we make it:
```json
{
   "outputs": {
     "o1": ...,
     "o2": ...
   },
   "FLYTE_DECK_FILE": ...
}
```

wdyt @kumare3, @hamersaw

EngHabu: Thinking about it again, DataCatalog already supports that leveling... 
Each ArtifactDataset can have multiple Artifacts attached to it.... today we create a DataSet per cacheKey (minus the inputs' hash) and then an Artifact attached to that per inputs' hash (i.e. a given task will have a single DataSet but for every invocation of that task, we will create a separate Artifact)
I think instead of adding the Deck URL as an additional output to the same artifact, we should, instead, add another Artifact to the same DataSet with a different tag (maybe `<inputs' hash>-flyte-deck` or something)
So you will basically have to call the `m.CreateArtifact()` a second time passing the `lt` variable you created here...
When we retrieve the artifacts [here](https://github.com/flyteorg/flytepropeller/blob/b9fb6e31eeb27a15c522995b0c0ea0f5536c0fe0/pkg/controller/nodes/task/catalog/datacatalog/datacatalog.go#L112-L122) we should then also try to retrieve the other artifact...

Pros:
* No naming collision, potential weird behavior when retrieving the artifacts

Cons:
* Additional API calls with potential of partial failures and added latency

pkg/controller/nodes/task/handler.go
```diff
@@ -541,6 +541,17 @@ func (t Handler) Handle(ctx context.Context, nCtx handler.NodeExecutionContext)
 				logger.Errorf(ctx, "failed to write cached value to datastore, err: %s", err.Error())
 				return handler.UnknownTransition, err
 			}
+			if r.GetDeckPath() != nil {
+				metadata, err := nCtx.DataStore().Head(ctx, *r.GetDeckPath())
+				if err == nil && metadata.Exists() {
+					tCtx.ow.GetDeckPath()
```
EngHabu, line 547: Not needed, right?

pkg/controller/nodes/transformers.go
```diff
@@ -79,7 +87,10 @@ func ToNodeExecutionEvent(nodeExecID *core.NodeExecutionIdentifier,
 	status v1alpha1.ExecutableNodeStatus,
 	eventVersion v1alpha1.EventVersion,
 	parentInfo executors.ImmutableParentInfo,
-	node v1alpha1.ExecutableNode, clusterID string, dynamicNodePhase v1alpha1.DynamicNodePhase) (*event.NodeExecutionEvent, error) {
+	node v1alpha1.ExecutableNode,
+	clusterID string,
+	dynamicNodePhase v1alpha1.DynamicNodePhase,
+	dataStore *storage.DataStore) (*event.NodeExecutionEvent, error) {
```
EngHabu, line 93: I don't think we should pass `dataStore` here...

Can you, instead, add DeckURI to [OutputInfo](https://github.com/flyteorg/flytepropeller/blob/59069d2aee79de353b2e1cc24c1912c89335faab/pkg/controller/nodes/handler/transition_info.go#L52). 
Then populate it wherever we populate OutputURI field?

EngHabu (commented):
Thank you, @pingsutw I think this is close!

pkg/controller/nodes/task/catalog/datacatalog/datacatalog.go
kumare3: Ohh absolutely not an output. I actually want this to be a meta output, remember, I think a deck is possible even for partial executions

EngHabu: @pingsutw, Maybe we drop it from catalog for now? this can happen later on when we figure out a good way of doing it?

pkg/controller/nodes/task/handler.go
pingsutw: yeah, we can remove it if we add deck_uri to output info

go.mod
pingsutw: updated it

pingsutw:
Few updates, cc @EngHabu 
- Upgrade flyteplugin
- Drop the changes in datacatalog (will create a new issue for caching deck)
- Add `deck_uri` in outputInfo

pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -491,3 +491,14 @@ func GetOutputsFile(outputDir DataReference) DataReference {
...
+
+func GetDeckFile(inputDir DataReference, store *storage.DataStore) DataReference {
+	deckURI := inputDir + "/deck.html"
+
+	// Deck file will not exist if we use the older version of flytekit or disable deck
+	metadata, err := store.Head(context.Background(), deckURI)
+	if err == nil && metadata.Exists() {
+		return deckURI
+	}
+	return ""
```
EngHabu, line 503: Let's not make any network calls here... the checks for the existence of the deck file should happen wherever we populate OutputUri (e.g. [here](https://github.com/flyteorg/flytepropeller/blob/59069d2aee79de353b2e1cc24c1912c89335faab/pkg/controller/nodes/handler/transition_info.go#L53))

pingsutw: Updated it

pingsutw:
cc @EngHabu

pingsutw:
I think it's ok to remove NodeStatus. IIUC, we will replace NodeStatus with OutputWriter. if so, we can use [OutputWriter.GetDeckPath()](https://github.com/flyteorg/flyteplugins/blob/f577f62727a16a282edfa7a49d985e47a4ccef79/go/tasks/pluginmachinery/ioutils/remote_file_output_writer.go#L38) to get `deck_uri`.

pingsutw:
Fixed it #453

PR #449, opened 2022-06-03
Calling abort rather than finalize on permanent failure

hamersaw:
# TL;DR
Currently FlytePropeller [calls "finalize"](https://github.com/flyteorg/flytepropeller/blob/b9fb6e31eeb27a15c522995b0c0ea0f5536c0fe0/pkg/controller/nodes/executor.go#L662) rather than "abort" on permanent failures. Consequently, during instances where a resource is actively running we fail to abort it. For example when the number of [timeouts](https://github.com/flyteorg/flytepropeller/blob/b9fb6e31eeb27a15c522995b0c0ea0f5536c0fe0/pkg/controller/nodes/executor.go#L389-L393) exceeds the maximum number of retries we [automatically transition](https://github.com/flyteorg/flytepropeller/blob/b9fb6e31eeb27a15c522995b0c0ea0f5536c0fe0/pkg/controller/nodes/executor.go#L397-L405) from a retryable failure to a permanent failure.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
I do not believe there are any unintended side-effects by calling abort. If a node has not yet been executed (ex. dynamic, subworkflow, etc) the abort should still be successful. If this is not the case, we will need to revisit this logic to ensure we fail correctly.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2298

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/nodes/executor.go
+++ b/pkg/controller/nodes/executor.go
@@ -659,7 +659,7 @@ func (c *nodeExecutor) handleNode(ctx context.Context, dag executors.DAGStructur
 
 	if currentPhase == v1alpha1.NodePhaseFailing {
 		logger.Debugf(ctx, "node failing")
-		if err := c.finalize(ctx, h, nCtx); err != nil {
+		if err := c.abort(ctx, h, nCtx, "node failing"); err != nil {
 			return executors.NodeStatusUndefined, err
 		}
 		nodeStatus.UpdatePhase(v1alpha1.NodePhaseFailed, v1.Now(), nodeStatus.GetMessage(), nodeStatus.GetExecutionError())
--- a/pkg/controller/nodes/executor_test.go
+++ b/pkg/controller/nodes/executor_test.go
@@ -790,6 +790,7 @@ func TestNodeExecutor_RecursiveNodeHandler_Recurse(t *testing.T) {
 				} else {
 					h.OnFinalizeMatch(mock.Anything, mock.Anything).Return(nil)
 				}
+				h.OnAbortMatch(mock.Anything, mock.Anything, mock.Anything).Return(nil)
 				hf.OnGetHandler(v1alpha1.NodeKindTask).Return(h, nil)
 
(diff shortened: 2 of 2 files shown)


pkg/controller/nodes/executor.go
```diff
@@ -659,7 +659,7 @@ func (c *nodeExecutor) handleNode(ctx context.Context, dag executors.DAGStructur
 
 	if currentPhase == v1alpha1.NodePhaseFailing {
 		logger.Debugf(ctx, "node failing")
-		if err := c.finalize(ctx, h, nCtx); err != nil {
```
EngHabu, line 662: Should we, instead, change finalize implementation in webapi to call delete?

Having both finalize and abort has always bothered me.. I think we only need one (perhaps finalize)... If you look at the implementations in all plugins, the code is pretty much the same... And when it isn't, it's usually a bug waiting to happen

hamersaw: I can certainly do that. I had two thoughts:
(1) The plugin abstraction, especially since users may write there own, makes it difficult to know if this issue exists in other plugins as well as the webapi async. On a permanent failure we should probably be attempting to abort anyways right? If we make the change in the webapi plugin are we potentially missing out on this fix in other plugins?
(2) I agree the two calls is confusing - especially when they are very similar. Since the existing [node-level abort implementation](https://github.com/flyteorg/flytepropeller/blob/b9fb6e31eeb27a15c522995b0c0ea0f5536c0fe0/pkg/controller/nodes/executor.go#L414-L425) calls both Abort and Finaliz if we introduce a Delete call in the webapi plugin are there any concerns about handling ResourceNotFound errors in webapi plugins? Just want to make sure we are not going to introduce downstream issues in custom plugins. This should already be handled right?

EngHabu (approved):
👍🏽

PR #450, opened 2022-06-08
Tracking terminated workflows in LRU cache

hamersaw:
# TL;DR
Adding a LRU filter cache to the resource version caching workflow store. We add the workflow resource version key if the workflow is successfully updated to a terminal phase. When workflows are requested with "get" we first lookup to see if the workflow has already been terminated. This fixes issues where we attempt to process a workflow that has already reached the terminal state, since the system is eventually consistent this may happen frequently.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2546

## Follow-up issue
_NA_


ckiosidis (commented):
FYI we did the exact same thing in an internal job that streams lineage data and it improved performance massively. 
We check the workflow status for success/failure and only cache those

EngHabu (approved):

PR #453, opened 2022-06-21
Decks proposal

EngHabu:
# TL;DR
Add deck_uri in NodeExecutionEvent. This change was originally proposed in https://github.com/flyteorg/flytepropeller/pull/443 but this is a modified (and simplified) implementation that only sends DeckUri for task nodes...

## Blocked By
https://github.com/flyteorg/flyteplugins/pull/272

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2175


hamersaw (approved):
Looks like a few lint issues to clean up. Otherwise LGTM!

pkg/controller/nodes/task/handler.go
```diff
@@ -479,7 +483,21 @@ func (t Handler) invokePlugin(ctx context.Context, p pluginCore.Plugin, tCtx *ta
 		if ee != nil {
 			pluginTrns.ObservedExecutionError(ee)
 		} else {
-			pluginTrns.ObserveSuccess(tCtx.ow.GetOutputPath(), tCtx.ow.GetDeckPath(), &event.TaskNodeMetadata{CacheStatus: cacheStatus.GetCacheStatus(), CatalogKey: cacheStatus.GetMetadata()})
+			var deckUri *storage.DataReference
+			exists, err := tCtx.ow.GetReader().DeckExists(ctx)
+			if err != nil {
+				logger.Errorf(ctx, "Failed to check deck file existence. Error: %v", err)
+				return pluginTrns, regErrors.Wrapf(err, "failed to check existence of deck file")
```
pingsutw, line 490: I think we don't need to return an error here because the deck file can be non-existent if people use an older version of flytekit or [disable flyte deck](https://github.com/flyteorg/flytekit/blob/27ac89732364fd96d767584fbda1af7696a41ac5/flytekit/core/task.py#L92) in flytekit.

pkg/controller/nodes/task/handler.go
```diff
@@ -522,26 +540,41 @@ func (t Handler) Handle(ctx context.Context, nCtx handler.NodeExecutionContext)
...
 				logger.Errorf(ctx, "failed to write cached value to datastore, err: %s", err.Error())
 				return handler.UnknownTransition, err
 			}
-			pluginTrns.CacheHit(tCtx.ow.GetOutputPath(), entry)
+
+			var deckUri *storage.DataReference
+			exists, err := r.DeckExists(ctx)
+			if err != nil {
+				logger.Errorf(ctx, "Failed to check deck file existence. Error: %v", err)
+				return handler.UnknownTransition, regErrors.Wrapf(err, "failed to check existence of deck file")
```
pingsutw, line 571: ditto

pkg/controller/nodes/task/handler.go
EngHabu: DeckExists should not return an error if the file doesn't exist. It should return a nil err and a false value for `exists` output... and that shouldn't be an error I agree.

An error here means we failed to even check the existence of the file, as in "s3 HEAD api call failed"... maybe an auth issue may be a rate limit issue... etc.

pkg/controller/nodes/task/handler.go
```diff
@@ -522,26 +540,41 @@ func (t Handler) Handle(ctx context.Context, nCtx handler.NodeExecutionContext)
...
+			} else if exists {
+				deckUriValue := tCtx.ow.GetDeckPath()
```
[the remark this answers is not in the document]
EngHabu: Aha, excellent observation! will fix that

pkg/controller/nodes/task/handler.go
pingsutw: I see, thanks

pkg/controller/nodes/task/handler.go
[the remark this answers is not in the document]
EngHabu: Removed it completely as we discussed earlier... people, for now, can click to follow the original execution of that task to find/access the decks... for now... to properly fix this, I think we need to modify the OutputReader interface in flyteplugins, specifically the Read() method to return OutputMetadata struct and the deckpath can be a field there...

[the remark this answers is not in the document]
pingsutw: Agreed, I can do that.

pingsutw (approved):

PR #461, opened 2022-07-21
Enable leader election on managed instances in a sharded deployment

hamersaw:
# TL;DR
This PR overrides the namespaced name of the leader election configuration for each managed instance. We set this to "propeller-leader-%d" where '%d' is the ID of the managed instance. All other leader election configuration relies on the PodTemplate used for managed instances. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
An alternative approach would be to parse the leader election configuration of the managed instance PodTemplate and append the ID onto that string rather than using the defaulted "propeller-leader" prefix. This approach gets far more complex than is necessary as we would have to parse config map mounts through the pod and into the volume map in the podtemplate container - ensuring we are correctly parsing overrides along the way. If hardcoding the leader election prefix becomes problematic in the future, we will have to take this approach.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2557

## Follow-up issue
_NA_

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -23,7 +23,6 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


EngHabu (approved):
Is there a way to specify that we want to run N number of replicas per instance? or this is just for the transient Node is being replaced scenarios?

hamersaw:
> Is there a way to specify that we want to run N number of replicas per instance? or this is just for the transient Node is being replaced scenarios?

So currently we're relying on the unique pod name to disallow duplicated propeller instances in sharded propeller. We manually disable leader election when creating managed propeller instances. However, we had a brief discussion about split-brain k8s clusters where leader election would be required, this PR goes a step further than relying on unique pod names and instead uses leader election for managed propeller instances. Still no support for multiple replicas per instance.

EngHabu (approved):

PR #468, opened 2022-08-09
Abort subworkflow on subnode failure

hamersaw:
# TL;DR
Currently, if an internal node in a subworkflow fails the parent workflow aborts. However, this abort does not traverse into the subworkflow to abort other nodes that may be running. This means that Flyte Pods may be left orphaned, this is especially important if finalizers are set and Flyte never cleans up the Pods.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
We updated the failure handling in the subworkflow to match the current dynamic task implementation. If an internal node (in the subworkflow) fails we mark the WorkflowNodeState as failing but keep the NodeState as running. Then in the next iteration FlytePropeller processes the failing subworkflow by first aborting it and then processing the failure node (if it ever exists). Errors are propagated up the workflow identically in failing subworkflows and dynamic tasks.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2533
fixes https://github.com/flyteorg/flyte/issues/2574

## Follow-up issue
_NA_


kumare3:
Ohh man this is a very good catch

honnix:
This is a very good finding! Unfortunately I do not have sufficient knowledge to do a constructive review, but I'm looking forward to testing it after it is merged.

ckiosidis:
Hey @hamersaw I can confirmed with some stuck Pods.
The pods are indeed orphans, the flyteworkflow k8s resources are not in the cluster and the executions finished with errors.

honnix:
Shall we get this going? It would be great to get this fix so we don't need to have a hacky way to clean up terminating pods. Thanks.

EngHabu (approved):

PR #479, opened 2022-09-02
Add structure attribute back to the CRD

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
https://github.com/flyteorg/flytepropeller/pull/442 remove `structure` from CRD at runtime, but we do need it for union type [validation](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/validators/typing.go#L33).
check [validateInputs](https://github.com/flyteorg/flytepropeller/blob/fd7b180accd1e15d9853806776d0b48c2afff03a/pkg/compiler/transformers/k8s/workflow.go#L199) -> [AreTypesCastable](https://github.com/flyteorg/flytepropeller/blob/0efd3affc54aa281789ed1463e304f28eee6bd81/pkg/compiler/transformers/k8s/inputs.go#L38) -> [CastsFrom](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/validators/typing.go#L19) -> [GetTagForType](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/validators/typing.go#L33)

Error:
```
  File "/Users/kevin/git/flytekit/flytekit/clients/raw.py", line 41, in handler
    return fn(*args, **kwargs)
  File "/Users/kevin/git/flytekit/flytekit/clients/raw.py", line 615, in create_execution
    return self._stub.CreateExecution(create_execution_request, metadata=self._metadata)
  File "/Users/kevin/opt/anaconda3/envs/flytekit-3.9/lib/python3.9/site-packages/grpc/_channel.py", line 946, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/Users/kevin/opt/anaconda3/envs/flytekit-3.9/lib/python3.9/site-packages/grpc/_channel.py", line 849, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.INTERNAL
	details = "Collected Errors: 2
	Error 0: Code: MismatchingTypes, Node Id: start-node, Description: Variable [b] (type [union_type:<variants:<simple:INTEGER > variants:<simple:STRING > > ]) doesn't match expected type [union_type:<variants:<simple:INTEGER structure:<tag:"int" > > > ].
	Error 1: Code: ParameterNotBound, Node Id: start-node, Description: Parameter not bound [b].
"
	debug_error_string = "{"created":"@1662108544.130630000","description":"Error received from peer ipv4:34.201.131.50:30081","file":"src/core/lib/surface/call.cc","file_line":1075,"grpc_message":"Collected Errors: 2\n\tError 0: Code: MismatchingTypes, Node Id: start-node, Description: Variable [b] (type [union_type:<variants:<simple:INTEGER > variants:<simple:STRING > > ]) doesn't match expected type [union_type:<variants:<simple:INTEGER structure:<tag:"int" > > > ].\n\tError 1: Code: ParameterNotBound, Node Id: start-node, Description: Parameter not bound [b].\n","grpc_status":13}"
```

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
The commands to repro, 
```
$ pip install git+https://github.com/flyteorg/flytekit@union-bug
$ pyflyte --config ~/.flyte/config-remote.yaml run --remote example_union1.py wf --b 2`
```
```
from flytekit import task, workflow                                                                           
from typing import Union, List
from enum import Enum


@task
def t1(a:  Union[int, str]) -> Union[int, str]:
    return "hello"

@task
def t2(a: Union[int, str]) -> Union[int, str]:
    return "hello"

@workflow
def wf(b:  Union[int, str]):
    t1(a=t2(a=b))
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2817
https://github.com/flyteorg/flytekit/pull/1144

## Follow-up issue
_NA_


hamersaw (changes requested):
@pingsutw So I think we should leave the stripping of the `Structure` field on type metadata, because this data is still only necessary at compile-time and will only increase the CRD size. Like you mentioned the [call to validateInputs](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/transformers/k8s/workflow.go#L192) may use the structure field in a corner case (kind of surprising that we just noticed this recently). It looks like we strip the metadata (as part of [this call](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/transformers/k8s/workflow.go#L169)) from the workflow interface before we check if the workflow inputs fit the interface. Can we instead rearrange these calls to validate the workflow inputs before the interface metadata is stripped?

pingsutw:
> Can we instead rearrange these calls to validate the workflow inputs before the interface metadata is stripped

yeah, agree. that makes sense. updating the pr

hamersaw (approved):

PR #482, opened 2022-09-12
Make k8s secrets optional

hamersaw:
# TL;DR
Currently, the volume mounts and env vars for secrets are not optional in pod definitions. This means that Pods created for Pods that are missing secrets can hang, and will remain in "Creating Container" status until the secret is created. By adding the optional flag to both volume mounts and env var secret injection allows progress if the secret does not exist. In the case of a Flyte task, this means flytekit will subsequently fail because the secret does not exist and in the progress provide an actionable error message, namely "... secret does not exist ..." rather than hanging.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2822

## Follow-up issue
_NA_


pkg/webhook/utils.go
```diff
@@ -51,6 +54,7 @@ func CreateVolumeForSecret(secret *core.Secret) corev1.Volume {
 						Path: strings.ToLower(secret.Key),
 					},
 				},
+				Optional: &optional,
```
EngHabu, line 57: While we are at it, can we make it read only as well?

EngHabu (approved):

pkg/webhook/utils.go
hamersaw: So when we actually [mount the volume](https://github.com/flyteorg/flytepropeller/blob/560bb1b0c198c2e7074e2f6aa74d45d933fbe846/pkg/webhook/utils.go#L62) it is already designated as ReadOnly. Does this cover your comment? It looks like on the `SecretVolumeSource ` there is a `DefaultMode *int32` field where we can use octal value (ie. 0555) to designate default mount permissions, but I'm not sure it's necessary.

EngHabu: Ah, we are good then!

PR #485, opened 2022-09-23
Move caching to node executor for fast cache hits

hamersaw:
# TL;DR
Currently cache lookups require 3 node-level phase changes, which equates to 3 streak rounds of flytepropeller execution. The overhead of updating the k8s FlyteWorkflow CRD after each of these rounds may make it seem like cache lookups are slow. This PR updates cache lookups to happen in the node executor, which allows us to transition nodes directly from NotYetStarted to Success and immediately process downstream nodes in the same FlytePropeller round making cache lookup overhead minimal.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
In addition to the performance improvements this PR refactors the code to support future implementation of caching other Flyte node types (ie. subworkflow / launchplan and array nodes when they are implemented).

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2886

## Follow-up issue
_NA_


pkg/controller/nodes/array/handler.go
```diff
@@ -295,6 +295,18 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 
 			retryAttempt := subNodeStatus.GetAttempts()
 
+			// fastcache will not emit task events for cache hits. we need to manually detect a
+			// transition to `SUCCEEDED` and add an `ExternalResourceInfo` for it.
+			if cacheStatus == idlcore.CatalogCacheStatus_CACHE_HIT && len(arrayEventRecorder.TaskEvents()) == 0 {
```
kumare3, line 300: is there one potential problem here, consider this case

Round 1:
  node 1 - cache-hit
   node 2 - cache-hit
   node 3 - system error
Round 2:
  node 1 - cache hit
  node 2 - cache miss
      - event write failure?

pkg/controller/nodes/array/handler.go
```diff
@@ -476,6 +488,7 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 			arrayNodeState.TaskPhaseVersion = taskPhaseVersion + 1
 		}
 
+		fmt.Printf("HAMERSAW - sending event with taskPhase %+v and version %d\n", taskPhase, taskPhaseVersion)
```
kumare3, line 491: remove this?

pkg/controller/nodes/array/handler.go
kumare3: Or maybe this will not happen, as when we actually get a system error I think we update the state?
is it possible to simulate this and test this?

One other option is allow admin to move the state from success to running!

kumare3 (commented):
I have one major comment, otherwise the code looks so much better.
This makes it actually way more readable. Not sure why the caching stuff was at the task level at all 🗡️

pkg/controller/nodes/array/handler.go
hamersaw: doh

kumare3 (approved):
thank you for the explanation

kumare3 (approved):

PR #498, opened 2022-11-02
Set default value for the input of FlyteTask

pingsutw:
# TL;DR

Failed to execute the FlyteTask because flyteadmin failed to marshal the data. Setting the default input to `core.LiteralMap{}` instead of nil can fix it.
```
    return _end_unary_response_blocking(state, call, False, None)
  File "/tmp/tmp.9GlVfLRofO/venv/lib/python3.10/site-packages/grpc/_channel.py", line 849, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.INTERNAL
        details = "failed to create workflow in propeller json: error calling MarshalJSON for type *v1alpha1.Inputs: Marshal called with nil"
        debug_error_string = "{"created":"@1666336057.217907913","description":"Error received from peer ipv4:127.0.0.1:30081","file":"src/core/lib/surface/call.cc","file_line":1074,"grpc_message":"failed to create workflow in propeller json: error calling MarshalJSON for type *v1alpha1.Inputs: Marshal called with nil","grpc_status":13}"
```

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3008

## Follow-up issue
_NA_

### DIFF
--- a/pkg/compiler/transformers/k8s/workflow.go
+++ b/pkg/compiler/transformers/k8s/workflow.go
@@ -178,6 +178,8 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 	} else if requiresInputs(wf) {
 		errs.Collect(errors.NewValueRequiredErr("root", "inputs"))
 		return nil, errs
+	} else {
+		inputs = &core.LiteralMap{}
 	}
 
 	for _, t := range tasks {
--- a/pkg/compiler/transformers/k8s/workflow_test.go
+++ b/pkg/compiler/transformers/k8s/workflow_test.go
@@ -117,6 +117,7 @@ func TestBuildFlyteWorkflow(t *testing.T) {
 	assert.Nil(t, wf.WorkflowSpec.Nodes[common.StartNodeID].Interruptible)
 	assert.Equal(t, "wf-1", wf.Labels[WorkflowNameLabel])
 	assert.Equal(t, "4", wf.Labels[ShardKeyLabel])
+	assert.NotNil(t, wf.Inputs)
 	assert.NoError(t, err)
 	assert.NotNil(t, wf)
 	errors.SetConfig(errors.Config{})


hamersaw:
@pingsutw this looks reasonable. My only question is how are workflows with no inputs handled when registering through pyflyte? It seems like they are probably defaulted to an empty LiteralMap? I don't feel strongly about whether this is handled flytekit-side or flyte-propeller-side just want a better understanding because we do not run into this issue with registered workflows.

pingsutw:
Actually we did wrap the empty input to [LiteralMap](https://github.com/flyteorg/flytekit/blob/b1dbab951e4d5562facf913e77b6d64c0da527e4/flytekit/remote/remote.py#L723), but seems like protibuf will convert it to Nil somehow if the value is empty. The reason why it works in flyteWorkflow is that we run [validation.CheckAndFetchInputsForExecution](https://github.com/flyteorg/flyteadmin/blob/2f2be8afddbd3261e58f20c42aed75535b644d47/pkg/manager/impl/execution_manager.go#L801-L805), and [it](https://github.com/flyteorg/flyteadmin/blob/2f2be8afddbd3261e58f20c42aed75535b644d47/pkg/manager/impl/validation/execution_validator.go#L79) will return empty literalMap

hamersaw:
@pingsutw when I try to run `pyflyte run` on a workflow with no input parameters I'm getting a pod error `tar: Removing leading `/' from member names`. Is that related to this or a separate issue?

pingsutw:
separate issue. I guess you run into this [issue](https://flyte-org.slack.com/archives/C043P7NDKL6/p1666911022385569), the pod is OOM, but flytekit doesn't show enough info about OOM.

pingsutw:
will fix it in flyteadmin side

PR #520, opened 2023-01-25
Abort tasks on non-success phase

hamersaw:
# TL;DR
FlytePropeller fast-fails when aborting tasks in a terminal phase. The result is that `Pods` marked as a retryable failure are never cleaned up. This PR only fast-fails when aborting tasks in a success phase, to ensure that all resources are correctly cleaned up at the risk of unnecessary API calls (ex. object does not exist, etc).

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently, the `TaskHandler` [fast-fails](https://github.com/flyteorg/flytepropeller/blob/20146304738db96e8a15cf9a824233676a7560ea/pkg/controller/nodes/task/handler.go#L773-L776) when aborting resources in a terminal phase. This causes issues where resources fail to cleanup when the task is reported as failed (ie. retryable failure or permanent failure) but the resource remains running.

One such scenario is handling `Pods` in the "ImagePullBackoff" state. We mark [these as failed](https://github.com/flyteorg/flyteplugins/blob/4634a81403e501882e3b1f39bcbd229b78768a4e/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L368-L372), but because of the fast-fail during abort these k8s resources are never deleted.

This may occur within other plugins as well. To be safe, we may want to attempt to abort all task phases other than `Success`.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3239

## Follow-up issue
_NA_


eapolinario (dismissed):

hamersaw:
Quick not on this - if we use this approach (ie. calling `Abort` even on terminal phases) then all failed pods will be automatically cleaned up by Flyte since `Abort` calls `Delete` on the object. This means that if a pod fails, we will be unable to manually view the pod to better understand the failure. This is required because currently things like 'ImagePullBackoff` for a long duration marks the task as a failure, but the pod is still actually running. So when we get to the `Abort` phase we can not be sure whether the pod is running or not. cc @kumare3 @EngHabu thoughts here?

flixr:
Anything I can help with here?
For me it would be better if the pods are not there anymore for debugging instead of getting relaunched by k8s in the background...

hamersaw:
closing in favor of https://github.com/flyteorg/flytepropeller/pull/541

PR #530, opened 2023-02-24
fix recovery of already recovered workflow

flixr:
# TL;DR
If an already recovered workflow was recovered again, all tasks were re-run.
This marks not only successful but also already recovered tasks as recoverable for futher runs.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

See slack discussion: https://flyte-org.slack.com/archives/CP2HDHKE1/p1677245731470789

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/nodes/executor.go
+++ b/pkg/controller/nodes/executor.go
@@ -190,6 +190,8 @@ func (c *nodeExecutor) attemptRecovery(ctx context.Context, nCtx handler.NodeExe
 	case core.NodeExecution_SKIPPED:
 		return handler.PhaseInfoSkip(nil, "node execution recovery indicated original node was skipped"), nil
 	case core.NodeExecution_SUCCEEDED:
+		fallthrough
+	case core.NodeExecution_RECOVERED:
 		logger.Debugf(ctx, "Node [%+v] can be recovered. Proceeding to copy inputs and outputs", nCtx.NodeExecutionMetadata().GetNodeExecutionID())
 	default:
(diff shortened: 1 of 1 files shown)


EngHabu (approved):
Thank you!

pkg/controller/nodes/executor.go
```diff
@@ -190,6 +190,7 @@ func (c *nodeExecutor) attemptRecovery(ctx context.Context, nCtx handler.NodeExe
 	case core.NodeExecution_SKIPPED:
 		return handler.PhaseInfoSkip(nil, "node execution recovery indicated original node was skipped"), nil
 	case core.NodeExecution_SUCCEEDED:
```
hamersaw, line 192: should add a `fallthrough` here right? in this case it will only print the debug if status is RECOVERED i believe.

flixr: Already out the door..
but I thought this would fall through just like in C. Sorry not fluent in go.. 😉

flixr: But yeah.. Makes sense that an explicit fallthrough is needed, otherwise it would not have worked for succeeded either..
feel free to fix it, otherwise I will on Monday

flixr: thanks for catching that, fixed.

hamersaw (approved):

PR #544, opened 2023-03-27
Remove resource injection on the node for container task

ByronHsu:
# TL;DR

Remove resource injection on the node for container task

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

In [BuildRawContainer](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L213), we do not inject the container resource from the  task template. Instead, we inject the resource from [parameters.TaskExecMetadata.GetOverrides().GetResources()](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L49).

This is because for both `_with_override(resource=...` and `@task(requests=...)`, they are both inject as the node resource override in [buildNodeSpec](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L49)

### Expected behavior

For the case of `@task(requests=...)`, the resource override on the node should be empty, and the resource is injected from tasktemplate at [BuildRawContainer](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L213)

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3540

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
 	github.com/fatih/color v1.13.0
 	github.com/flyteorg/flyteidl v1.3.14
-	github.com/flyteorg/flyteplugins v1.0.48
+	github.com/flyteorg/flyteplugins v1.0.47
 	github.com/flyteorg/flytestdlib v1.0.15
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-redis/redis v6.15.7+incompatible
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
 	github.com/fatih/color v1.13.0
 	github.com/flyteorg/flyteidl v1.3.14
-	github.com/flyteorg/flyteplugins v1.0.45
+	github.com/flyteorg/flyteplugins v1.0.48
 	github.com/flyteorg/flytestdlib v1.0.15
 	github.com/ghodss/yaml v1.0.0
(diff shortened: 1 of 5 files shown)


EngHabu:
I'm probably missing something but should we be trying to do that here? https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L305

What issue are you seeing, @ByronHsu ?

ByronHsu:
@EngHabu Thanks for the quick reply. Yes, I have another PR injecting the resource there. See the attachment in the PR https://github.com/flyteorg/flyte/issues/3540.
No issue for the OSS version, but we are hacking something internally, and this logic flaw is blocking us.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
 	github.com/fatih/color v1.13.0
 	github.com/flyteorg/flyteidl v1.3.14
-	github.com/flyteorg/flyteplugins v1.0.45
+	github.com/flyteorg/flyteplugins v1.0.48
```
hamersaw, line 10: this should be 1.0.47 - https://github.com/flyteorg/flyteplugins/releases. 1.0.48 will pull in code deprecating the barrier transitions, which is not meant to be a part of this merge.

ByronHsu: done. thanks for the catch

hamersaw (approved):

PR #547, opened 2023-03-30
Add support for GCP secret manager

jeevb:
Tested with:
```
from flytekit import Secret, task, workflow


@task(secret_requests=[Secret(group="mainsecret", key="na", group_version="1")])
def secret_task() -> str:
    with open("/etc/flyte/secrets/mainsecret/1") as f:
        return f.read()


@workflow
def wf() -> str:
    x = secret_task()
    return x
```

Screenshot 2023-03-30 at 2 03 25 PM

Note the use of the required `key`, even though the secrets plugin does not use it. @wild-endeavor: we may want to consider making this optional. Flytekit will also need support for including the group version [in the file path](https://github.com/flyteorg/flytekit/blob/6c52297ae65914bec5988e6fbae0a20fa9917599/flytekit/core/context_manager.py#L356) when fetching secrets.

### DIFF
--- a/pkg/webhook/gcp_secret_manager_test.go
+++ b/pkg/webhook/gcp_secret_manager_test.go
@@ -1,0 +1,79 @@
+package webhook
+
+import (
+	"context"
+	"testing"
+
+	"github.com/flyteorg/flytepropeller/pkg/webhook/config"
+
+	"github.com/go-test/deep"
+
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestGCPSecretManagerInjector_Inject(t *testing.T) {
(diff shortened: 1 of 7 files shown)


wild-endeavor (approved):

pkg/webhook/gcp_secret_manager_test.go
```diff
@@ -0,0 +1,79 @@
...
+			},
+
+			InitContainers: []corev1.Container{
+				{
+					Name:  "gcp-pull-secret-0",
+					Image: "gcr.io/google.com/cloudsdktool/cloud-sdk:alpine",
+					Command: []string{
+						"sh",
+						"-c",
+						"gcloud secrets versions access 2 --secret=TestSecret --out-file=/etc/flyte/secrets/TestSecret/2 && chmod +rX /etc/flyte/secrets/TestSecret /etc/flyte/secrets/TestSecret/2",
```
wild-endeavor, line 43: maybe lower path?

EngHabu (dismissed):
🥳 awesome! thank you, @jeevb

wild-endeavor (approved):

PR #550, opened 2023-04-06
Instrument ArrayNode

hamersaw:
# TL;DR
Instrumenting support for ArrayNodes to provide a functionally complete map task experience.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1131

## Follow-up issue
_NA_


pkg/controller/nodes/array/execution_context.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strconv"
+
+	"github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
+	"github.com/flyteorg/flytepropeller/pkg/controller/executors"
+)
+
+const (
+	FlyteK8sArrayIndexVarName string = "FLYTE_K8S_ARRAY_INDEX"
```
kumare3, line 11: would have loved for this and next to be a shorter string
I infact added the shorter strings in flytekit, but never ended up making changes in propeller. this is to reduce the size of the pod object itself

pkg/controller/nodes/array/execution_context.go
```diff
@@ -0,0 +1,48 @@
+package array
```
kumare3, line 1: how about a comment explaining why we need a special context. As i do not think we have a special context in any other node type?

pkg/controller/nodes/array/handler.go
```diff
@@ -0,0 +1,593 @@
...
+	externalResources := make([]*event.ExternalResourceInfo, 0, len(arrayNodeState.SubNodePhases.GetItems()))
+	messageCollector := errorcollector.NewErrorMessageCollector()
+	switch arrayNodeState.Phase {
+	case v1alpha1.ArrayNodePhaseExecuting, v1alpha1.ArrayNodePhaseFailing:
+		currentParallelism := uint32(0)
+		for i, nodePhaseUint64 := range arrayNodeState.SubNodePhases.GetItems() {
+			nodePhase := v1alpha1.NodePhase(nodePhaseUint64)
+
+			// do not process nodes that have not started or are in a terminal state
+			if nodePhase == v1alpha1.NodePhaseNotYetStarted || isTerminalNodePhase(nodePhase) {
```
kumare3, line 82: should we also continue in case the node is already in a terminal state?

pkg/controller/nodes/array/handler.go
```diff
@@ -0,0 +1,593 @@
...
+
+			// create array contexts
+			arrayNodeExecutor, arrayExecutionContext, arrayDAGStructure, arrayNodeLookup, subNodeSpec, subNodeStatus, arrayEventRecorder, err :=
+				a.buildArrayNodeContext(ctx, nCtx, &arrayNodeState, arrayNode, i, &currentParallelism)
+			if err != nil {
+				return handler.UnknownTransition, err
+			}
+
+			// execute subNode
+			_, err = arrayNodeExecutor.RecursiveNodeHandler(ctx, arrayExecutionContext, arrayDAGStructure, arrayNodeLookup, subNodeSpec)
```
kumare3, line 275: the event stuff seems to be repeated, but i guess for array nodes it is slightly different right?

pkg/controller/nodes/array/node_lookup.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (a *arrayNodeLookup) GetNodeExecutionStatus(ctx context.Context, id v1alpha1.NodeID) v1alpha1.ExecutableNodeStatus {
+	if id == a.subNodeID {
+		return a.subNodeStatus
+	}
+
+	return a.NodeLookup.GetNodeExecutionStatus(ctx, id)
+}
+
+func newArrayNodeLookup(nodeLookup executors.NodeLookup, subNodeID v1alpha1.NodeID, subNodeSpec *v1alpha1.NodeSpec, subNodeStatus *v1alpha1.NodeStatus) arrayNodeLookup {
```
kumare3, line 33: this is interesting, the same pattern has just been followed. makes it easier to read if we know the core types right

kumare3 (commented):
To be honest this is a crazy big code change and it looks pretty good. Boy though, when you look at some of the repeated verbosity in golang haha. But, mostly looks great to me, couple small comments.

Since this is behind a feature gate, i think this is safe to go

pkg/controller/nodes/array/handler.go
hamersaw: This should be covered by the `isTerminalNodePhase` check here.

pkg/controller/nodes/array/handler.go
hamersaw: exactly, there is a little bit of verbosity to keep this as a drop-in replacement. basically al this eventing stuff just captures everything that happens when evaluating subnodes and then aggregates the information into our `ExternalResources` field, which is how maptasks currently report everything.

I talked to the FE team briefly, for now I implemented this as a drop-in replacement, but maybe before GA we clean up this representation. very open to discussion here, but for now the goal was to make this as compatible as possible and provide a superset of functionality.

pkg/controller/nodes/array/handler.go
```diff
@@ -0,0 +1,593 @@
...
+
+	messageCollector := errorcollector.NewErrorMessageCollector()
+	switch arrayNodeState.Phase {
+	case v1alpha1.ArrayNodePhaseExecuting, v1alpha1.ArrayNodePhaseFailing, v1alpha1.ArrayNodePhaseSucceeding:
+		currentParallelism := uint32(0)
+		for i, nodePhaseUint64 := range arrayNodeState.SubNodePhases.GetItems() {
+			nodePhase := v1alpha1.NodePhase(nodePhaseUint64)
+
+			// do not process nodes that have not started
+			if nodePhase == v1alpha1.NodePhaseNotYetStarted {
```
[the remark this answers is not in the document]
hamersaw: i think you're probably right - added it.

pkg/controller/nodes/array/execution_context.go
hamersaw: great idea, added it on the `buildArrayNodeContext` function in the handler code.

kumare3 (approved):

PR #556, opened 2023-04-17
Implement ability to specify additional/override annotations when using Vault Secret Manager

pradithya:
# TL;DR
Add `annotations` configuration to the Vault Secret Manager to provide better flexibility when integrating with Vault.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

When the Vault secret manager is enabled, flyte-pod-webhook is responsible for adding vault-specific annotation to the task pod that requests secret variables. These annotations are used to configure the Vault Agent behavior when retrieving and populating the secret from Vault. As of now, the annotation provided by `flyte-pod-webhook` is limited to the one implemented in `vault_secret_manager.go`. 

This PR allows Flyte administrator to add additional or override default annotations providing better flexibility to customize Vault agent behavior. For example, the following configuration will set the Vault agent [authentication type](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-auth-config) to use GCP  and to immediately exit if the authentication failed.

```
      vaultSecretManager:
        kvVersion: "2"
        role: flyte
        annotations:
          vault.hashicorp.com/auth-config-type : "gce"
          vault.hashicorp.com/auth-type : "gcp"
          vault.hashicorp.com/agent-auto-auth-exit-on-err : "true"
```

## Tracking Issue
_NA_

## Follow-up issue
_NA_


pkg/webhook/vault_secret_manager.go
```diff
@@ -74,8 +74,9 @@ func (i VaultSecretManagerInjector) Inject(ctx context.Context, secret *coreIdl.
 			return p, false, err
 		}
 
-		p.ObjectMeta.Annotations = utils.UnionMaps(p.ObjectMeta.Annotations, commonVaultAnnotations)
-		p.ObjectMeta.Annotations = utils.UnionMaps(p.ObjectMeta.Annotations, secretVaultAnnotations)
+		p.ObjectMeta.Annotations = utils.UnionMaps(i.cfg.ExtraAnnotations, p.ObjectMeta.Annotations)
```
pradithya, line 77: Note that I change the ordering of `utils.UnionMaps` arguments so that user-defined annotations are given priority. This allows overriding annotation created by the webhook.
For instance, in my org, we have a multi-tenant Vault where we need a specific `role` to access a secret. However, Flyte <> Vault integration can only support single role. Thus, to use different role user can do so by specifying annotation in their workflow
```
@task(
    secret_requests=[
        Secret(
            group=SECRET_GROUP,
            key=SECRET_NAME,
            mount_requirement=Secret.MountType.FILE,
        )
    ],
    pod_template=PodTemplate(annotations={
        "vault.hashicorp.com/role" : ROLE,
    })
)
```

kumare3:
Cc @EngHabu / @hamersaw

pkg/webhook/config/config.go
```diff
@@ -113,8 +113,9 @@ type GCPSecretManagerConfig struct {
 }
 
 type VaultSecretManagerConfig struct {
-	Role      string    `json:"role" pflag:",Specifies the vault role to use"`
-	KVVersion KVVersion `json:"kvVersion" pflag:"-,The KV Engine Version. Defaults to 2. Use 1 for unversioned secrets. Refer to - https://www.vaultproject.io/docs/secrets/kv#kv-secrets-engine."`
+	Role             string            `json:"role" pflag:",Specifies the vault role to use"`
+	KVVersion        KVVersion         `json:"kvVersion" pflag:"-,The KV Engine Version. Defaults to 2. Use 1 for unversioned secrets. Refer to - https://www.vaultproject.io/docs/secrets/kv#kv-secrets-engine."`
+	ExtraAnnotations map[string]string `json:"extraAnnotations" pflag:"-,Additional annotation to be added to the pod. Useful to further customize Vault integration (https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations)"`
```
hamersaw, line 118: Is the term `extraAnnotations` more descriptive than simply `annotations`?

pkg/webhook/vault_secret_manager_test.go
```diff
@@ -137,6 +157,18 @@ func TestVaultSecretManagerInjector_Inject(t *testing.T) {
 			want:    ExpectedKVv2,
 			wantErr: false,
 		},
+		{
+			name: "KVv3 Secret - with extra annotations",
+			args: args{
+				cfg: config.VaultSecretManagerConfig{Role: "flyte", KVVersion: config.KVVersion2, ExtraAnnotations: map[string]string{
```
hamersaw, line 163: Can you add some more examples here to cover the various override scenarios?

hamersaw (commented):
This is great! I think the only point of contention is ensuring users understand the default vault-specific annotations that Flyte sets so they don't assume this `extraAnnotations` argument needs to contain something already being set. Does it make sense to update [these docs](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/containerization/use_secrets.html#vault-secrets-manager) as part of this change as well?

pkg/webhook/config/config.go
pradithya: Make sense, I think `annotations` is the better term since it not only allows Flyte administrator to add extra annotations but also overrides the default annotation created by Flyte.

pkg/webhook/vault_secret_manager_test.go
pradithya: Added cases where users provide their annotation via pod template annotations and system override annotation.

pradithya:
> Does it make sense to update [these docs](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/containerization/use_secrets.html#vault-secrets-manager) as part of this change as well?

Yes, will add a follow-up PR to update the Vault-related user guide once the PR is merged.

pkg/webhook/vault_secret_manager.go
```diff
@@ -74,8 +74,9 @@ func (i VaultSecretManagerInjector) Inject(ctx context.Context, secret *coreIdl.
 			return p, false, err
 		}
 
-		p.ObjectMeta.Annotations = utils.UnionMaps(p.ObjectMeta.Annotations, commonVaultAnnotations)
-		p.ObjectMeta.Annotations = utils.UnionMaps(p.ObjectMeta.Annotations, secretVaultAnnotations)
+		p.ObjectMeta.Annotations = utils.UnionMaps(i.cfg.Annotations, p.ObjectMeta.Annotations)
+		p.ObjectMeta.Annotations = utils.UnionMaps(commonVaultAnnotations, p.ObjectMeta.Annotations)
+		p.ObjectMeta.Annotations = utils.UnionMaps(secretVaultAnnotations, p.ObjectMeta.Annotations)
```
hamersaw, line 79: IIUC `UnionMaps` takes multiple arguments and the later are favored. So this can otherwise be:
```
p.ObjectMeta.Annotations = utils.UnionMaps(secretVaultAnnotations, commonVaultAnnotations, i.cfg.Annotations, p.ObjectMeta.Annotations)
```
I think it makes this a little easier to understand.

hamersaw (dismissed):
Minor nit, otherwise looks great.

pkg/webhook/vault_secret_manager.go
pradithya: Neat, didn't realize it's a variadic function. Updated in my latest commit.

hamersaw (approved):

bimtauer:
Discovered this now when making some updates to my original contribution. I love this feature and its great to know that someone out there is also using flyte with vault :) One thing: I will rename the tests - you introduced a bunch of new ones named KVv3, KVv4 and so on. This was meant to represent the vault kv secret backend version to be tested. To my knowledge there are only 2.

kumare3:
Tbh @bimtauer i atleast know 3/4 companies that use vault

PR #573, opened 2023-06-07
Use GetExecutionData instead

EngHabu:
# TL;DR
Removes deprecated code path to retrieve LP's outputs and moves to the separate GetExecutionData() call.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3753

### DIFF
--- a/pkg/controller/nodes/subworkflow/launchplan/admin.go
+++ b/pkg/controller/nodes/subworkflow/launchplan/admin.go
@@ -45,6 +45,7 @@ type executionCacheItem struct {
 	core.WorkflowExecutionIdentifier
 	ExecutionClosure *admin.ExecutionClosure
 	SyncError        error
+	ExecutionOutputs *core.LiteralMap
 }
 
 func (e executionCacheItem) ID() string {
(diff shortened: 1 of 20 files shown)


pkg/controller/nodes/subworkflow/launchplan/admin.go
```diff
@@ -246,12 +247,30 @@ func (a *adminLaunchPlanExecutor) syncItem(ctx context.Context, batch cache.Batc
 			continue
 		}
 
+		execData, err := a.adminClient.GetExecutionData(ctx, &admin.WorkflowExecutionGetDataRequest{
```
katrogan, line 250: For performance reasons, would it be worthwhile to wrap this call in an if block and only fetch execution data when [RawOutputPolicy](https://github.com/flyteorg/flytepropeller/blob/eadf07ba3ac2e54ee4aaf95a98c9941a5a005646/pkg/controller/config/config.go#L246) is set to inline?

katrogan (approved):
Would we want to similarly propage

pkg/controller/nodes/subworkflow/launchplan/admin.go
EngHabu: Discussed offline. We always need to call GetExecutionData()

PR #592, opened 2023-07-19
Check for None values in branch nodes

pingsutw:
# TL;DR
blocked by https://github.com/flyteorg/flytekit/pull/1747
Add support comparison (EQ and NEQ) between nil and non-nil values in the branch nodes. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
import random
import typing
from flytekit import conditional, task, workflow


@task
def coin_toss(seed: int) -> typing.Optional[bool]:
    r = random.Random(seed)
    if r.random() < 0.5:
        return True
    return None


@task
def failed() -> int:
    return -1


@task
def success() -> int:
    return 0


@workflow
def wf(seed: int = 5) -> int:
    result = coin_toss(seed=seed)
    return conditional("test").if_(result.is_none()).then(success()).else_().then(failed())


if __name__ == '__main__':
    wf()
```

image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3514

## Follow-up issue
_NA_


pkg/controller/nodes/branch/evaluator.go
```diff
@@ -24,10 +24,14 @@ func EvaluateComparison(expr *core.ComparisonExpression, nodeInputs *core.Litera
 	var rPrim *core.Primitive
 
 	if expr.GetLeftValue().GetPrimitive() == nil {
-		if nodeInputs == nil {
-			return false, errors.Errorf(ErrorCodeMalformedBranch, "Failed to find Value for Variable [%v]", expr.GetLeftValue().GetVar())
+		if len(expr.GetLeftValue().GetVar()) == 0 {
```
hamersaw, line 27: So we're representing a `None` value by using a variable of length 0? and then automatically converting it into a scalar with `NoneValue`. Would it make more sense to set this as a scalar with `NoneValue` from flytekit?

pkg/controller/nodes/branch/comparator.go
```diff
@@ -116,24 +126,32 @@ func Evaluate(lValue *core.Primitive, rValue *core.Primitive, op core.Comparison
...
+			return false, errors.Errorf(ErrorCodeMalformedBranch, "Only primitives can be compared. LHS Variable is non primitive.")
+		}
 	}
 	return Evaluate(lValue.GetScalar().GetPrimitive(), rValue, op)
 }
 
 func EvaluateLiterals(lValue *core.Literal, rValue *core.Literal, op core.ComparisonExpression_Operator) (bool, error) {
 	if lValue.GetScalar() == nil || lValue.GetScalar().GetPrimitive() == nil {
-		return false, errors.Errorf(ErrorCodeMalformedBranch, "Only primitives can be compared. LHS Variable is non primitive.")
+		if lValue.GetScalar().GetNoneType() == nil {
```
hamersaw, line 147: Any reason why all of these are on a separate line rather than just being included in the same `if` conjunction?

pkg/controller/nodes/branch/evaluator.go
pingsutw: > Would it make more sense to set this as a scalar with NoneValue from flytekit?

Yes, we can, however, the idl must be changed since none is not in the primitives. we can add `is_none bool` to the [operand](https://github.com/flyteorg/flyteidl/blob/95e11cca2dac18b727f122bbc4456ea6ab499289/protos/flyteidl/core/condition.proto#L30), wdyt?

hamersaw: Is your intuition that we have `Primitive` rather than `Scalar` in the operand because this was meant to be very simple?

I think I would support adding a `Void none_type` to the operand [similar to this](https://github.com/flyteorg/flyteidl/blob/95e11cca2dac18b727f122bbc4456ea6ab499289/protos/flyteidl/core/literals.proto#L86).

pkg/controller/nodes/branch/comparator.go
pingsutw: updated it

pkg/controller/nodes/branch/evaluator.go
pingsutw: updated Idl. https://github.com/flyteorg/flyteidl/pull/427

hamersaw:
@pingsutw can we update here? looks like https://github.com/flyteorg/flyteidl/pull/427 was merged.

pingsutw:
Sorry @hamersaw, forget this PR, could you +1? We've merged flytekit and flyteidl PRs

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4154. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

PR #596, opened 2023-07-26
Adding flags for ignore-retry-cause and default-max-attempts

hamersaw:
# TL;DR
Adding flags to simplify retry behavior. `ignore-retry-cause` means that system and user failures are treated the same and `default-max-attempts` sets a default number of retries for tasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_


hamersaw:
In the current implementation enabling `ignoreRetryCause` also turns the `interruptibleFailureThreshold` configuration into a complementary value. IMO this will result in unintended behavior and should be handled separately.

pkg/controller/nodes/node_exec_context.go
```diff
@@ -289,19 +289,21 @@ func (c *nodeExecutor) BuildNodeExecutionContext(ctx context.Context, executionC
...
-		maxAttempts := uint32(config.GetConfig().NodeConfig.DefaultMaxAttempts)
+		currentAttempt := int32(s.GetAttempts() + s.GetSystemFailures())
+		maxAttempts := int32(config.GetConfig().NodeConfig.DefaultMaxAttempts)
 		if n.GetRetryStrategy() != nil && n.GetRetryStrategy().MinAttempts != nil && *n.GetRetryStrategy().MinAttempts != 0 {
-			maxAttempts = uint32(*n.GetRetryStrategy().MinAttempts)
+			maxAttempts = int32(*n.GetRetryStrategy().MinAttempts)
 		}
 
-		if interruptible && currentAttempt >= maxAttempts-c.interruptibleFailureThreshold {
+		if interruptible && ((c.interruptibleFailureThreshold > 0 && currentAttempt >= c.interruptibleFailureThreshold) ||
```
fg91, line 298: If we understand correctly, one can specify a positive and a negative number for both the old and the new behaviour. Is there a place in flytepropeller where the config is sanity checked and rules can be enforced early on?
If yes, what do you think about simply requiring a negative number in the config if the new behaviour is chosen and a positive one in the other case?
This would make the logic here simpler.

Also, what do you think of moving the checks whether the retry is interruptible into two separate functions with descriptive names? This would make the if conditions here easier to understand. Also would make this specific logic easier to unit test.

pkg/controller/nodes/node_exec_context.go
```diff
@@ -286,10 +294,25 @@ func (c *nodeExecutor) BuildNodeExecutionContext(ctx context.Context, executionC
...
-	// a node is not considered interruptible if the system failures have exceeded the configured threshold
-	if interruptible && s.GetSystemFailures() >= c.interruptibleFailureThreshold {
-		interruptible = false
-		c.metrics.InterruptedThresholdHit.Inc(ctx)
+	if config.GetConfig().NodeConfig.IgnoreRetryCause {
+		// For the unified retry behavior we execute the last interruptibleFailureThreshold attempts on a non
+		// interruptible machine
+		maxAttempts := uint32(config.GetConfig().NodeConfig.DefaultMaxAttempts)
+		if n.GetRetryStrategy() != nil && n.GetRetryStrategy().MinAttempts != nil && *n.GetRetryStrategy().MinAttempts != 1 {
+			maxAttempts = uint32(*n.GetRetryStrategy().MinAttempts)
```
pingsutw, line 302: Should we deprecate `MinAttempts` and add `MaxAttempts` 😂

pingsutw (approved):

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4153. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

PR #605, opened 2023-08-09
Add support namespace override

pingsutw:
# TL;DR
- Instead of setting the namespace for the k8s resource every time, we set it only when the k8s resource's namespace is None
- Only add an owner reference if the task's namespace matches the Flyteworkflow's.
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3898

## Follow-up issue
_NA_

### DIFF
--- a/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -102,9 +102,7 @@ type PluginManager struct {
 }
 
 func (e *PluginManager) AddObjectMetadata(taskCtx pluginsCore.TaskExecutionMetadata, o client.Object, cfg *config.K8sPluginConfig) {
-	if len(o.GetNamespace()) == 0 {
-		o.SetNamespace(taskCtx.GetNamespace())
-	}
+	o.SetNamespace(taskCtx.GetNamespace())
 	o.SetAnnotations(utils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), utils.CopyMap(taskCtx.GetAnnotations())))
 	o.SetLabels(utils.UnionMaps(cfg.DefaultLabels, o.GetLabels(), utils.CopyMap(taskCtx.GetLabels())))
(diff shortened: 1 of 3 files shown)


pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -102,12 +102,15 @@ type PluginManager struct {
 }
 
 func (e *PluginManager) AddObjectMetadata(taskCtx pluginsCore.TaskExecutionMetadata, o client.Object, cfg *config.K8sPluginConfig) {
-	o.SetNamespace(taskCtx.GetNamespace())
+	if len(o.GetNamespace()) == 0 {
```
pingsutw, line 105: cc @hamersaw Is there any concern to do this?

pingsutw: Maybe make it opt-in. like
```
if !K8sPluginConfig.namespaceOverride || len(o.GetNamespace()) == 0 {
    o.SetNamespace(taskCtx.GetNamespace())
}
```

hamersaw: This seems very hacky to me. I think the next obvious ask is going to be to support this for all plugins. So rather than adding an option to each config object (ie. Ray, Dask, Spark, Pod, etc) is there a way we can add this override to the task execution context so that it's applicable to all k8s plugin types?

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4152. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

PR #615, opened 2023-09-11
[flytepropeller] Support attribute access on promises

ByronHsu:
See https://github.com/flyteorg/flyteidl/pull/439

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3864

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -146,4 +146,3 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
-replace github.com/flyteorg/flyteidl => ../flyteidl
--- a/go.mod
+++ b/go.mod
@@ -146,3 +146,4 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
+replace github.com/flyteorg/flyteidl => ../flyteidl
--- a/pkg/compiler/validators/bindings.go
+++ b/pkg/compiler/validators/bindings.go
@@ -125,6 +125,12 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 					}
 				}
 
+				// Skip the validation if the promise has attribute paths
+				// because we don't know the type of the resolved attribute
+				if len(val.Promise.AttrPath) > 0 {
+					return param.GetType(), []c.NodeID{val.Promise.NodeId}, true
+				}
+
 				if !validateParamTypes || AreTypesCastable(sourceType, expectedType) {
 					val.Promise.NodeId = upNode.GetId()
 					return param.GetType(), []c.NodeID{val.Promise.NodeId}, true
--- a/pkg/compiler/validators/bindings.go
+++ b/pkg/compiler/validators/bindings.go
@@ -125,10 +125,24 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 					}
 				}
 
-				// Skip the validation if the promise has attribute paths
-				// because we don't know the type of the resolved attribute
-				if len(val.Promise.AttrPath) > 0 {
+				// If the type is a struct (e.g. dataclass) and the attribute path is longer than 0,
+				// We skip the type check and let it fail at runtime because we don't know the type of struct field
+				if sourceType.GetSimple() == flyte.SimpleType_STRUCT && len(val.Promise.AttrPath) > 0 {
 					return param.GetType(), []c.NodeID{val.Promise.NodeId}, true
+				}
+
+				// If the variable has an attribute path. Extract the type of the last attribute.
+				for range val.Promise.AttrPath {
+					if sourceType.GetCollectionType() != nil {
+						sourceType = sourceType.GetCollectionType()
+					}
+					if sourceType.GetMapValueType() != nil {
+						sourceType = sourceType.GetMapValueType()
+					}
+					// If the current type is struct, skip the type check because we don't know the type of struct field
+					if sourceType.GetSimple() == flyte.SimpleType_STRUCT {
+						return param.GetType(), []c.NodeID{val.Promise.NodeId}, true
+					}
 				}
 
 				if !validateParamTypes || AreTypesCastable(sourceType, expectedType) {
--- a/pkg/controller/nodes/attr_path_resolver.go
+++ b/pkg/controller/nodes/attr_path_resolver.go
@@ -1,0 +1,149 @@
+package nodes
+
+import (
+	"context"
+
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/flyteorg/flytepropeller/pkg/controller/nodes/errors"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// resolveAttrPathInPromise resolves the literal with attribute path
+// If the promise is chained with attributes (e.g. promise.a["b"][0]), then we need to resolve the promise
+func resolveAttrPathInPromise(ctx context.Context, nodeID string, literal *core.Literal, bindAttrPath []*core.PromiseAtrribute) (*core.Literal, error) {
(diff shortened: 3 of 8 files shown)


pkg/compiler/validators/bindings.go
```diff
@@ -125,6 +125,12 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 					}
 				}
 
+				// Skip the validation if the promise has attribute paths
+				// because we don't know the type of the resolved attribute
+				if len(val.Promise.AttrPath) > 0 {
```
pingsutw, line 130: we should be able to know the attribute type if the promise type is list or dict, right?

pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -0,0 +1,149 @@
...
+				Value: &core.Scalar_Generic{
+					Generic: new_st,
+				},
+			},
+		}
+	case []interface{}:
+		literals := []*core.Literal{}
+		for _, v := range currVal.([]interface{}) {
+			literalItem := &core.Literal{}
+			scalar, err := resolveInterfaceToLiteralScalar(ctx, nodeID, v)
```
pingsutw, line 102: does it work if the current value is a nested list? [[1, 2, 3]]

pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -0,0 +1,149 @@
...
+	value := &core.Primitive{}
+
+	switch obj.(type) {
+	case string:
+		value.Value = &core.Primitive_StringValue{StringValue: obj.(string)}
+	case int:
+		value.Value = &core.Primitive_Integer{Integer: int64(obj.(int))}
+	case float64:
+		value.Value = &core.Primitive_FloatValue{FloatValue: obj.(float64)}
+	case bool:
```
pingsutw, line 136: How about the other types in the scalar? https://github.com/flyteorg/flyteidl/blob/7f62a157921db31d5f7a2285b4e9a42b4a01eaa4/protos/flyteidl/core/literals.proto#L84-L90

pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -0,0 +1,149 @@
...
+
+	var currVal interface{}
+	var tmpVal interface{}
+	var exist bool
+
+	currVal = st.AsMap()
+
+	// Turn the current value to a map so it can be resolved more easily
+	for _, attr := range bindAttrPath {
+		switch currVal.(type) {
```
pingsutw, line 64: should we use `reflect.TypeOf` instead? defer to @hamersaw

pkg/controller/nodes/attr_path_resolver.go
ByronHsu: brilliant! I made it recursive to handle nested cases and added a test case for that

pkg/compiler/validators/bindings.go
hamersaw: We should be able to use the [AreTypesCastable function](https://github.com/flyteorg/flytepropeller/blob/6b1d11c736c4a9d8b8fa48264de0d3dc5853d494/pkg/compiler/validators/typing.go#L358) here right? It should handle `Union` as either a source or destination type. IMO this feature needs type validation, it's something Flyte is very opinionated about and seems very sloppy to leave out.

pkg/controller/nodes/attr_path_resolver.go
hamersaw: I don't feel strongly either way.

pkg/controller/nodes/output_resolver.go
```diff
@@ -54,11 +54,19 @@ func (r remoteFileOutputResolver) ExtractOutput(ctx context.Context, nl executor
 		actualVar = variable
 	}
 
+	var output *core.Literal
```
hamersaw, line 57: maybe add some comments here to make this more readable? For example:
```
// retrieving task output
...
// resolving binding attribute paths (if exist)
```
I found this much more difficult to follow than I would have expected at first look.

hamersaw (commented):
Looks close to me! Great work.

pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return currVal, nil
+}
+
+// resolveAttrPathInPbStruct resolves the protobuf struct (e.g. dataclass) with attribute path
+func resolveAttrPathInPbStruct(ctx context.Context, nodeID string, st *structpb.Struct, bindAttrPath []*core.PromiseAttribute) (*core.Literal, error) {
```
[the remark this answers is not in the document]
ByronHsu: The idea is similar but the impl is different. `resolveAttrPathInPromise` is to resolve literal, and `resolveAttrPathInPbStruct` is to resolve protobuf struct

pkg/controller/nodes/output_resolver.go
ByronHsu: added

pkg/compiler/validators/bindings.go
```diff
@@ -125,6 +125,26 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 					}
 				}
 
+				// If the type is a struct (e.g. dataclass) and the attribute path is longer than 0,
+				// We skip the type check and let it fail at runtime because we don't know the type of struct field
+				if sourceType.GetSimple() == flyte.SimpleType_STRUCT && len(val.Promise.AttrPath) > 0 {
```
hamersaw, line 130: @wild-endeavor is this necessarily true. if we have `dataclass_json` do we know the type?

go.mod
```diff
@@ -146,3 +146,4 @@ require (
 )
 
 replace github.com/aws/amazon-sagemaker-operator-for-k8s => github.com/aws/amazon-sagemaker-operator-for-k8s v1.0.1-0.20210303003444-0fb33b1fd49d
+replace github.com/flyteorg/flyteidl => ../flyteidl
```
hamersaw, line 149: @ByronHsu can you update this?

pkg/compiler/validators/bindings.go
wild-endeavor: @ByronHsu would you mind just trying to fill this in a bit?  If it's really not doable, or if it has a bunch of edge cases, I think it's okay.

Walking through some examples.  Let's say the simple case of
```python
@task
def t1() -> List[str]:
    ...

@task
def t2(needs_str: str):
    ...

@workflow
def wf():
    res = t1()
    t2(needs_str=res[5])
```
In this case `t1` will have a TypedInterface output of
```
        value {
          type {
            collection_type {
              simple: STRING
            }
          }
          description: "o0"
        }
```

and because it's a string it will type check against the string in `t2`.  The same can be applied for nested lists and dictionaries.  As long as they are expressed via `collection_type` and `map_value_type` they are relatively easily type-checkable. I assume this is what the code below is doing?


The dataclass case is more complicated
```python
@dataclass_json
@dataclass
class MyDC(object):
    snapshotDate: datetime
    region: str

@task
def t1(needs_dt: datetime):
    ...

@task
def t2(needs_str: str):
    ...

@workflow
def wf(a: MyDC):
    t1(needs_dt=a.snapshotDate)
    t2(needs_str=a.region)
```

The reason it's more complicated is because the dataclass types are completely obscured (esp. since flyte idl currently doesn't support multi-variate map types).  I assume this is why you're skipping checking in the simple/struct case.

Could you see if it's possible though to capture it?  Can we
* Add a new field to LiteralType that is only relevant for the scalar case.  Maybe just in `LiteralType` or in `TypeStructure`.
* Add a nested literal map of the types that's only present in the dataclass case.  Just iterate through the fields in dataclass and recursively call the TypeEngine.
* Add the same checking logic in propeller as the normal `map_value_type` if it's a simple_struct and this new field is present.

What do you think?  It will add to the correctness of this new feature.  And it will make Dan happy.  And in the end, that's what we're all really about.

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4150. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

